PR #123, opened 2026-04-09
Fix wrap main advice

martyall:


packages/pickles/src/Pickles/Types.purs
```diff
@@ -774,3 +789,279 @@ instance
...
+        fs
+    in
+      uncurry2
+        ( \unfinalizedProofs messagesForNextStepProof ->
+            WrapPrevProofState { unfinalizedProofs, messagesForNextStepProof }
+        )
+        tup
+
+instance
+  ( CheckedType f c (Vector mpv (PerProofUnfinalized WrapIPARounds sfvar fvar bvar))
```
martyall: The correct constraint would be on `PerProofUnfinalized WrapIPARounds sfvar fvar bvar)` and inherit the Vector constraint for free (maybe you need Reflectable mpv Int)

packages/pickles/src/Pickles/Types.purs
```diff
@@ -774,3 +789,279 @@ instance
...
+    let
+      tup :: Tuple2 (Vector slot0Width (Vector WrapIPARounds var)) (Vector slot1Width (Vector WrapIPARounds var))
+      tup = genericFieldsToVar
+        @(Tuple2 (Vector slot0Width (Vector WrapIPARounds a)) (Vector slot1Width (Vector WrapIPARounds a)))
+        fs
+    in
+      uncurry2 (\slot0 slot1 -> WrapOldBpChals { slot0, slot1 }) tup
+
+instance
+  ( CheckedType f c (Vector slot0Width (Vector WrapIPARounds var))
```
martyall: The correct constraint would be on var and inherit this constraint through nested vector for free

packages/pickles/src/Pickles/Types.purs
```diff
@@ -774,3 +789,279 @@ instance
...
+  , CheckedType f c (Vector slot1Width (Vector WrapIPARounds var))
```
martyall: ditto

PR #124, opened 2026-04-17
Step wrap ml

martyall:


packages/pickles/src/Pickles/Linearization/Env.purs
```diff
@@ -190,7 +154,7 @@ precomputeAlphaPowers maxPow alpha = label "precompute-alpha-powers" $ go 2 [ co
   go i acc
     | i > maxPow = pure acc
     | otherwise = do
-        let prev = unsafePartial $ fromJust $ Array.last acc
+        let prev = fromJust' "precomputeAlphaPowers: `acc` seeded with [1, alpha], never empty" (Array.last acc)
```
martyall: this could be defined using NonEmtpty and avoid the `fromJust'`  as the base case is NonEmpty

packages/pickles/src/Pickles/Linearization/Env.purs
martyall, line 198: can `alphaPowers` be non-empty allowing us to drop the `fromJust'`

martyall:
It would seem that many of the modules in the PlonkChecks namespace are now pretty trivial, maybe containing one binding and some type defs. Would it be possible to drop the nesting and just include these in a single PlonkChecks module?

packages/pickles/src/Pickles/Proof/Dummy.purs
```diff
@@ -0,0 +1,116 @@
...
+  -> Proof Pallas.G Vesta.BaseField
+dummyWrapProof bcd =
+  let
+    prf = bcd.proofDummy
+    evals = bcd.dummyEvals
+
+    -- Pallas generator g0 = Tock.Curve.(to_affine_exn one). Pallas points
+    -- have coordinates in Pallas.BaseField = Vesta.ScalarField.
+    g0 :: AffinePoint Vesta.ScalarField
+    g0 = fromJust'
```
martyall: We happen to know that generator is Affine so the fromJust is safe here -- the only non-affine point is the identity element which is clearly not a generator

packages/pickles/src/Pickles/Prove/Pure/Common.purs
```diff
@@ -0,0 +1,464 @@
...
+-- |   `ft_eval0 = permContribution - pEval0Folded - constantTerm`
+-- |
+-- | where
+-- |
+-- | * `permContribution` = `term1 - term2 + boundary` from
+-- |   `Pickles.PlonkChecks.Permutation.permContribution`;
+-- | * `pEval0Folded` = Horner fold of `pEval0Chunks` at
+-- |   `zeta^(2^srsLengthLog2)`;
+-- | * `constantTerm` = `evaluate linearizationPoly (fieldEnv ...)`.
+ftEval0
```
martyall, line 401: we have another ftEval0 which is way dumber. why?

packages/pickles/src/Pickles/Prove/Pure/Step.purs
```diff
@@ -0,0 +1,916 @@
...
+    -- Step 3: expand previous-proof bp challenges (used both for the
+    -- challenges_digest sub-sponge and for combined_inner_product's
+    -- b_poly inputs).
+    prevExpanded :: Vector n (Vector d StepField)
+    prevExpanded = map (map expandChal) input.oldBulletproofChallenges
+
+    -- Step 4a: sub-sponge over flattened expanded prev challenges.
+    challengesDigest :: StepField
+    challengesDigest = evalPureSpongeM initialSponge do
+      for_ prevExpanded \chals -> for_ chals (absorb :: StepField -> _)
```
martyall: is this annotation needed. if so, can we switch to visible type application ?

packages/pickles/src/Pickles/Prove/Pure/Wrap.purs
```diff
@@ -0,0 +1,550 @@
...
+    -- expands them and evaluates `b_poly(zeta) + r·b_poly(zetaw)`.
+    rawPrechalsArray :: Array StepField
+    rawPrechalsArray = pallasProofOpeningPrechallenges input.verifierIndex
+      { proof: input.proof
+      , publicInput: input.publicInput
+      , prevChallenges: prevChallengeList
+      }
+
+    rawPrechalsVec :: Vector StepIPARounds (SizedF 128 StepField)
+    rawPrechalsVec = fromJust'
```
martyall: We should move this check somewhere in the FFI. sually the FFI is structured into a type class and you're rarely making the rust call directly. Maybe it can happen as part of the type class method

martyall:
i feel like there is a lot of manual wrapping unwrapping of `F` constructor. The whole point of the `F` constructor is that anything that is pure field related that is happening inside a circuit is probably going to be in F, and unless yo uhave some good reason like an external dependency, you will keep it in F if you can. F tries to pass along any of the field related type classes and as many other low level type classes as posssible, so probably any "pure" field function can be written using F. If the problem is coming from the FFI, the place to do the wrapping/unwrapping might be as part of the FFI definition

packages/pickles/src/Pickles/Prove/Pure/Wrap.purs
```diff
@@ -0,0 +1,550 @@
...
+        :< UnChecked (crossFieldSized128 dv.xi)
+        :< Vector.nil
+
+    -- ===== 3 digests (all wrap field) =====
+    --
+    -- Order matches OCaml's `to_data`:
+    --   (sponge_digest, msg_for_next_wrap, msg_for_next_step)
+    digests :: Vector 3 (F WrapField)
+    digests =
+      F (crossFieldDigest dv.spongeDigestBeforeEvaluations)
```
martyall: could also map F

packages/pickles/src/Pickles/Prove/Pure/Wrap.purs
```diff
@@ -0,0 +1,550 @@
...
+    branchData :: F WrapField
+    branchData = F (packBranchDataWrap dv.branchData)
+
+    -- ===== Feature flag + lookup slots — constant zeros =====
+    --
+    -- OCaml's `Spec.T.Constant` in `wrap_packed_typ` still allocates
+    -- these as field elements (with the check skipped). For
+    -- `Features.Full.none` + `lookup.use = No`, all slots are zero.
+    featureFlags :: Vector 8 (F WrapField)
+    featureFlags = Vector.replicate (F zero)
```
martyall: I think (F f) is also a field, so zero @(F WrapField) would also work

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+  -> StepAdvice prevsSpec StepIPARounds WrapIPARounds inputVal len carrier
+buildStepAdvice input =
+  let
+    -- Pallas generator (= OCaml `Tock.Curve.one`). Reused for every
+    -- curve-point field in the base-case dummy advice.
+    g0 :: AffinePoint (F StepField)
+    g0 = coerce
+      ( fromJust'
+          "Pallas generator to affine (never the identity element)"
+          (Curves.toAffine (Curves.generator :: Pallas.G))
```
martyall: As noted elsewhere, for the generator this operation is completely safe. so can just drop the baggage

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+    --     = Vector.init most_recent_width (fun _ -> Dummy.Ipa.Wrap.sg)
+    -- We reify the runtime width to the type level so
+    -- `hashMessagesForNextStepProofPure` gets a `Vector n` of the
+    -- correct length.
+    msgStepDigestStepField :: StepField
+    msgStepDigestStepField = case input.mostRecentWidth of
+      0 -> hashMessagesForNextStepProofPure
+        { stepVk: wrapVkStep
+        , appState: valueToFields @StepField @inputVal input.prevPublicInput
+        , proofs: (Vector.nil :: Vector 0 _)
```
martyall: i doubt you need this annotation

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+
+    -- `prevProofs` has `mostRecentWidth` entries, matching OCaml
+    -- `proof.ml:168-171`:
+    --   messages_for_next_step_proof.challenge_polynomial_commitments
+    --     = Vector.init most_recent_width (fun _ -> Dummy.Ipa.Wrap.sg)
+    -- We reify the runtime width to the type level so
+    -- `hashMessagesForNextStepProofPure` gets a `Vector n` of the
+    -- correct length.
+    msgStepDigestStepField :: StepField
+    msgStepDigestStepField = case input.mostRecentWidth of
```
martyall: can we use `Finite 3` here for `mostRecentWitdth` and make it clear from the type

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+      0 -> hashMessagesForNextStepProofPure
+        { stepVk: wrapVkStep
+        , appState: valueToFields @StepField @inputVal input.prevPublicInput
+        , proofs: (Vector.nil :: Vector 0 _)
+        }
+      1 -> hashMessagesForNextStepProofPure
+        { stepVk: wrapVkStep
+        , appState: valueToFields @StepField @inputVal input.prevPublicInput
+        , proofs: singleEntry :< (Vector.nil :: Vector 0 _)
```
martyall: i doubt you need the annotation. Also isn't there a vector singleton? maybe there should be

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+        }
+      2 -> hashMessagesForNextStepProofPure
+        { stepVk: wrapVkStep
+        , appState: valueToFields @StepField @inputVal input.prevPublicInput
+        , proofs: singleEntry :< singleEntry :< (Vector.nil :: Vector 0 _)
```
martyall: replicate?

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+    -- [spongeDigest, msgWrap, msgStep]. `fopProofState.spongeDigest`
+    -- is already a step-field value; coerce to wrap bits.
+    sponge0 :: F StepField
+    sponge0 = fop.spongeDigestBeforeEvaluations
+
+    digests3 :: Array WrapField
+    digests3 =
+      [ stepToWrap sponge0
+      , msgWrapDigestWrapField -- already wrap field
+      , Curves.fromBigInt (Curves.toBigInt msgStepDigestStepField)
```
martyall: didn't we make a special method just for digests? Rather than calling unsafeCoerce via BigInt directly?

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+-- | StepField)). The wrap VK's commitments are Pallas points with
+-- | coordinates in Pallas.BaseField = StepField, so no cross-field
+-- | coercion is needed.
+extractWrapVKCommsAdvice
+  :: VerifierIndex PallasG WrapField
+  -> VerificationKey (WeierstrassAffinePoint PallasG (F StepField))
+extractWrapVKCommsAdvice vk =
+  let
+    raw :: Array (AffinePoint StepField)
+    raw = vestaVerifierIndexColumnComms vk
```
martyall: if this is Array because of the FFI, maybe convering to the right length vector should be done at a lower level (with throw), so that we can use safe static vector ops below

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+-- | `hashMessagesForNextStepProofPure` in the step field — the dummy
+-- | wrap proof's `messages_for_next_step_proof` hash mirrors OCaml's
+-- | `Common.hash_messages_for_next_step_proof` on the real wrap VK.
+extractWrapVKForStepHash
+  :: VerifierIndex PallasG WrapField
+  -> StepVK StepField
+extractWrapVKForStepHash vk =
+  let
+    raw :: Array (AffinePoint StepField)
+    raw = vestaVerifierIndexColumnComms vk
```
martyall: if this is Array because of the FFI, maybe convering to the right length vector should be done at a lower level (with throw), so that we can use safe static vector ops below

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+    -- for domainLog2 ∈ [0,15], mask bits). The mask follows OCaml's
+    -- `ones_vector ~first_zero:most_recent_width |> Vector.rev`
+    -- (wrap_main.ml:231-238):
+    --   N0 → [F, F] → packed = 0
+    --   N1 → [F, T] → packed = 2
+    --   N2 → [T, T] → packed = 3
+    packedBranchData :: WrapField
+    packedBranchData =
+      let
+        maskBits = case input.mostRecentWidth of
```
martyall: if using `Finite 3` this would be safe

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+     , msgWrapDigest :: WrapField
+     -- | Pre-computed FOP proof state to serialize. Picks which Ro
+     -- | state's plonk values to use (SimpleChain vs Tree_proof_return
+     -- | vs module-init). Caller passes e.g.
+     -- | `Dummy.simpleChainStepDummyFopProofState { proofsVerified }` or
+     -- | `Dummy.treeStepDummyFopProofState { proofsVerified }`.
+     , fopProofState ::
+         UnfinalizedProof StepIPARounds (F StepField) (Type1 (F StepField)) Boolean
+     }
+  -> Array WrapField
```
martyall, line 517: Why not return a Vector here?

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+-- | `wrapCompile` threads the advice to `runWrapProverT` only for
+-- | type-class instance resolution; the values are never read during
+-- | circuit shape walking. Callers hand this in to `wrapCompile` when
+-- | they don't yet have real advice (e.g. when they need the compiled
+-- | wrap VK BEFORE the step solver runs so they can feed it to
+-- | `buildStepAdviceWithOracles`).
+-- |
+-- | Polymorphic in the slot shape via `PadSlots.replicateSlots`, which
+-- | builds a `slots`-shaped value of zero bp-challenge stacks.
+zeroWrapAdvice
```
martyall: so this is wrong, I'm seeing `wrapCompile` uses 

```purescript
compileAction
      :: WrapProverT branches mpv slots m
           (CircuitBuilderState (KimchiGate WrapField) (AuxState WrapField))
```

But actually what you shoul dhave is an `Effect` based instance/interpreter for the advice monad that just throws an error (usually with a string reflecting the advice requested, maybe `WrapCircuitBuilder`. and you will run compile in that. That way you would still get an error if this happened, but you know it never will because that's just teh way it works -- circuit building never uses advice

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+       carrierVar
+  => CheckedType StepField (KimchiConstraint StepField) input
+  => Monad m
+  => StepProveContext len
+  -> StepRule len inputVal input outputVal output prevInputVal prevInput
+  -> StepAdvice prevsSpec StepIPARounds WrapIPARounds inputVal len carrier
+  -> m StepCompileResult
+stepCompile ctx rule advice = do
+  let
+    compileAction
```
martyall: I dont think you want to run the compileAction with `StepProverT`, it will force you to create some BS advice value

But actually what you should have is an Effect based instance/interpreter for the advice monad that just throws an error (usually with a string reflecting the advice requested, maybe WrapCircuitBuilder. and you will run compile in that. That way you would still get an error if this happened, but you know it never will because that's just teh way it works -- circuit building never uses advice

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+       StepIPARounds
+       WrapIPARounds
+       (FVar StepField)
+       (Type2 (SplitField (FVar StepField) (BoolVar StepField)))
+       (BoolVar StepField)
+       len
+       carrierVar
+  => CheckedType StepField (KimchiConstraint StepField) input
+  => Monad m
+  => (Error -> m (StepProveResult outputSize))
```
martyall: Why not just return an ExceptT ProverError m ?

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+        csSatisfied = verifyProverIndex @StepField @VestaG
+          { proverIndex: compileResult.proverIndex, witness, publicInputs }
+      in
+        if not csSatisfied then do
+          let
+            _ = unsafePerformEffect $
+              dumpRowLabels
+                (Array.length compileResult.builtState.publicInputs)
+                compileResult.builtState.constraints
+          onError (error "stepProve: constraint system not satisfied (wrote row→label map to /tmp/ps_step_row_labels.txt)")
```
martyall: when using Excpet monad, you can just unsafeThrow this. This is an instrumentation that will be gone in the final version anyway

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+  => Add mpv 45 totalBases
+  => Add 1 totalBasesPred totalBases
+  => PadSlots slots mpv
+  => CircuitType WrapField
+       (slots (Vector WrapIPARounds (F WrapField)))
+       (slots (Vector WrapIPARounds (FVar WrapField)))
+  => CheckedType WrapField (KimchiConstraint WrapField)
+       (slots (Vector WrapIPARounds (FVar WrapField)))
+  => Monad m
+  => (Error -> m WrapProveResult)
```
martyall: its better to reutrn in ExceptT ProverError monad

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+                  in
+                    { row: row + nRows, out: out <> [ line ] }
+              )
+              { row: 0, out: [] }
+              compileResult.builtState.constraints
+          FS.writeTextFile UTF8 "/tmp/ps_wrap_row_labels.txt"
+            (Array.intercalate "\n" out <> "\n")
+      in
+        if not csSatisfied then
+          onError (error "wrapProve: constraint system not satisfied")
```
martyall: This is part of the prover debug mode only, we already have a mechanism for this.

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+    Left e -> onError (error ("stepProve solver: " <> show e))
+    Right (Tuple publicOutputs assignments) ->
+      let
+        { witness, publicInputs } = makeWitness
+          { assignments
+          , constraints: map _.variables compileResult.constraints
+          , publicInputs: compileResult.builtState.publicInputs
+          }
+
+        csSatisfied = verifyProverIndex @StepField @VestaG
```
martyall: This should probably not be running every time by default, but rather in some debug mode. We should guard all this extra checking and instrumentation behind a debug check

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+  -> m WrapProveResult
+wrapSolveAndProve onError ctx compileResult = do
+  let
+    rawSolver
+      :: SolverT WrapField (KimchiConstraint WrapField)
+           (WrapProverT branches mpv slots m)
+           (WrapStatementPacked StepIPARounds (Type1 (F WrapField)) (F WrapField) Boolean)
+           Unit
+    rawSolver =
+      makeSolver' (emptyProverState { debug = true }) (Proxy @(KimchiConstraint WrapField))
```
martyall: we dont want to run in debug mode all the time, this should be an option

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+      let
+        { witness, publicInputs } = makeWitness
+          { assignments
+          , constraints: map _.variables compileResult.constraints
+          , publicInputs: compileResult.builtState.publicInputs
+          }
+
+        csSatisfied = verifyProverIndex @WrapField @PallasG
+          { proverIndex: compileResult.proverIndex, witness, publicInputs }
+        _dumpRowLabels = unsafePerformEffect do
```
martyall: put this in a debug mode guarded block

packages/pickles/src/Pickles/Prove/Step.purs
```diff
@@ -0,0 +1,2001 @@
...
+      :: SolverT StepField (KimchiConstraint StepField)
+           ( StepProverT prevsSpec StepIPARounds WrapIPARounds inputVal
+               len
+               carrier
+               m
+           )
+           Unit
+           (Vector outputSize (F StepField))
+    rawSolver =
+      makeSolver' (emptyProverState { debug = true })
```
martyall: debug option should be passed to the top level function and inherited here

packages/pickles/src/Pickles/Prove/Wrap.purs
```diff
@@ -0,0 +1,821 @@
...
+              , assignments
+              }
+
+extractStepVKComms
+  :: VerifierIndex VestaG StepField
+  -> StepVK WrapField
+extractStepVKComms vk =
+  let
+    raw :: Array (AffinePoint WrapField)
+    raw = pallasVerifierIndexColumnComms vk
```
martyall: if the FFI handles the conversion to Vector, we can use safe ops below

packages/pickles/src/Pickles/Step/FinalizeOtherProof.purs
```diff
@@ -323,7 +325,7 @@ finalizeOtherProofCircuit ops params { unfinalized, witness, mask, prevChallenge
   zetaToNMinus1 <- domainVanishingPoly domainWhich zeta params.domainLog2
 
   let
-    alphaPow n = unsafePartial $ fromJust $ Array.index alphaPowers n
+    alphaPow n = fromJust' ("step.FinalizeOtherProof: alpha^" <> show n <> " (precomputeAlphaPowers output)") $ Array.index alphaPowers n
```
martyall: can we use  a static vector here for safe indexing

packages/pickles/src/Pickles/Wrap/FinalizeOtherProof.purs
```diff
@@ -233,7 +241,7 @@ wrapFinalizeOtherProofCircuit params vanishingPolynomial { unfinalized, witness,
     vanishingPolynomial zeta
 
   let
-    alphaPow n = unsafePartial $ fromJust $ Array.index alphaPowers n
+    alphaPow n = fromJust' ("wrap.FinalizeOtherProof: alpha^" <> show n <> " (precomputeAlphaPowers output)") $ Array.index alphaPowers n
```
martyall: is there a way to use vector for owers of alpha to make indexing safe?

packages/pickles/src/Pickles/Wrap/Main.purs
```diff
@@ -179,24 +196,84 @@ stepAllEvalsToProofWitness (StepAllEvals r) =
...
+  => FopBodyParams WrapField
+  -> Int -- slotIdx, for label only
+  -> PlonkDomain WrapField t m
+  -> UnfinalizedView
+  -> ProofWitness (FVar WrapField)
+  -> Vector PaddedLength (Vector WrapIPARounds (FVar WrapField)) -- pre-padded chals
+  -> Snarky (KimchiConstraint WrapField) t m (Vector WrapIPARounds (FVar WrapField))
+processOneSlotFopBody fopBaseParams slotIdx domain unfView witness paddedChals = do
+  { finalized, expandedChallenges } <- wrapFinalizeOtherProofCircuit
+    ( Record.merge
```
martyall, line 249: I think you can actually just use record update syntax here rather than Recrod.merge?

packages/pickles/src/Pickles/Wrap/Main.purs
```diff
@@ -395,82 +531,104 @@ wrapMain config (WrapStatementPacked stmtR) = do
...
+    let
+      revIdxs :: Vector mpv Int
+      revIdxs = Vector.reverse (Vector.generate @mpv getFinite)
+      revWdis = Vector.reverse wrapDomainIndices
+      revInputs = Vector.zip revIdxs revWdis
+    revDomains <- traverse
+      ( \(Tuple slotIdx wdi) -> do
+          which <- label ("block3-wrap-domain-" <> show slotIdx) $
+            Pseudo.oneHotVector @3 wdi
+          Pseudo.toDomain @16 domainConfig which config.allPossibleDomainLog2s
```
martyall, line 571: what is this magic 16 number?

packages/pickles/src/Pickles/Wrap/Main.purs
```diff
@@ -395,82 +531,104 @@ wrapMain config (WrapStatementPacked stmtR) = do
...
+  domains <- do
+    let
+      revIdxs :: Vector mpv Int
+      revIdxs = Vector.reverse (Vector.generate @mpv getFinite)
+      revWdis = Vector.reverse wrapDomainIndices
+      revInputs = Vector.zip revIdxs revWdis
+    revDomains <- traverse
+      ( \(Tuple slotIdx wdi) -> do
+          which <- label ("block3-wrap-domain-" <> show slotIdx) $
+            Pseudo.oneHotVector @3 wdi
```
martyall, line 569: what is this magic 3 number

packages/pickles/src/Pickles/Dummy.purs
```diff
@@ -133,205 +144,277 @@ replicateChal
...
+    do
+      -- Evals record RTL: selectors first, then sigma, z, coefficients, w, public_input, ft_eval1
+      idxEndomulScalar <- pointEval
+      idxEmul <- pointEval
+      idxMul <- pointEval
+      idxCompleteAdd <- pointEval
+      idxPoseidon <- pointEval
+      idxGeneric <- pointEval
+      let
+        indexEvals = fromJust' "dummyEvals indexEvals: literal array of 6 selectors" $ Vector.toVector @6
```
martyall: use :< to construct it

packages/pickles/src/Pickles/Dummy.purs
```diff
@@ -112,7 +123,7 @@ chal = do
   ro <- get
   let next = ro.chalCounter + 1
   put $ ro { chalCounter = next }
-  pure $ fromBits $ unsafePartial fromJust $ Vector.toVector @128 $ bitsRandomOracle 128 ("chal_" <> show next)
+  pure $ fromBits $ fromJust' "Dummy chal: `bitsRandomOracle 128` yields exactly 128 bits" $ Vector.toVector @128 $ bitsRandomOracle 128 ("chal_" <> show next)
```
martyall: can bitsRandomOracle return a vector?

packages/pickles/src/Pickles/Dummy.purs
```diff
@@ -133,205 +144,277 @@ replicateChal
 replicateChal = do
   let n = reflectType (Proxy @n)
   arr <- sequence (Array.replicate n (chal :: RoM (SizedF 128 f)))
```
martyall: does Vector have replicate?

packages/pickles/src/Pickles/Dummy.purs
```diff
@@ -133,205 +144,277 @@ replicateChal
...
+  in
+    do
+      idxEndomulScalar <- pointEval
+      idxEmul <- pointEval
+      idxMul <- pointEval
+      idxCompleteAdd <- pointEval
+      idxPoseidon <- pointEval
+      idxGeneric <- pointEval
+      let
+        indexEvals = fromJust' "proofDummyPrevEvals indexEvals: literal array of 6 selectors" $ Vector.toVector @6
```
martyall: construct by hand using :<

packages/pickles/src/Pickles/Dummy.purs
```diff
@@ -406,44 +479,48 @@ computeDummySgValues pallasSrs vestaSrs =
...
-        , xiRaw: unsafePartial fromJust $ SizedF.fromField @128
-            (Curves.fromBigInt (BigInt.fromInt 1 + BigInt.pow (BigInt.fromInt 2) (BigInt.fromInt 64)) :: WrapField)
+        { alphaRaw: u.plonk.alpha
+        , betaRaw: u.plonk.beta
+        , gammaRaw: u.plonk.gamma
+        , zetaRaw: u.plonk.zeta
+        , xiRaw: fromJust'
+            "Dummy xiRaw: dummy-challenge `1 + 2^64` fits in 128 bits"
+            ( SizedF.fromField @128
+                (Curves.fromBigInt (BigInt.fromInt 1 + BigInt.pow (BigInt.fromInt 2) (BigInt.fromInt 64)) :: WrapField)
```
martyall: if you use fromInt, use homomorphic property of BigIntFromInt and will have less noise

packages/pickles/src/Pickles/Sponge.purs
```diff
@@ -323,20 +307,14 @@ lowest128Bits' constrainLowBits endo x = do
...
       lo :: SizedF 128 (F f)
-      lo =
-        unsafePartial fromJust
-          $ SizedF.fromField @128
-          $ fromBigInt
-          $
-            mod xBig two128
+      lo = fromJust'
+        "Sponge squeezeScalar lo: `mod xBig 2^128` always fits 128 bits"
+        (SizedF.fromField @128 (fromBigInt (mod xBig two128)))
```
martyall: can just use unsafePartial fromJust here, it's clear from the callsite

packages/pickles/src/Pickles/Sponge.purs
```diff
@@ -361,7 +339,7 @@ lowest128BitsPure x =
     two128 = BigInt.pow (BigInt.fromInt 2) (BigInt.fromInt 128)
     lo = fromBigInt (mod xBig two128)
   in
-    unsafePartial fromJust $ fromField @128 lo
+    fromJust' "lowest128BitsPure: `mod xBig 2^128` always fits 128 bits" (fromField @128 lo)
```
martyall: it's clear this is safe from the callsite

packages/snarky/src/Snarky/Circuit/Types.purs
```diff
@@ -243,6 +245,44 @@ instance
...
+-- | `Const Unit a` carries no content regardless of `a`, so its
+-- | `CircuitType` instance is trivial (zero wires, empty arrays).
+-- | The `a` on the RHS is a fresh type variable the fundep resolves
+-- | to whatever the caller needs.
+-- |
+-- | This instance, together with the `Product` instance below, lets
+-- | higher-kinded "functor-product slot lists" (see
+-- | `Pickles.Wrap.Slots`) be allocated by `exists` without a custom
+-- | newtype wrapper.
+instance CircuitType f (Const Unit a) (Const Unit avar) where
```
martyall: You could probably get a more general instance here where yo uhave a constraint (CircuitType f a avar => CircuitType f (Const a b) (Const avar bvar)

FOLLOW-UPS.md
```diff
@@ -46,3 +46,42 @@ The `ones_vector` logic in `WrapMain.purs` (block1) computes the `actual_proofs_
 ## assertAll_ is defined but uncalled
 
 `Snarky.Circuit.DSL.Assert.assertAll_` matches OCaml's `Boolean.Assert.all` (void assertion, sum-based). It's implemented and exported but not yet called by any circuit code. It exists for completeness — the OCaml equivalent is used in some circuit paths we haven't translated yet.
+
```
martyall, line 66: I think this whole file needs to be updated

PR #137, opened 2026-05-27
Proof ffi to kimchi

martyall:


packages/snarky-kimchi/src/Snarky/Backend/Kimchi/Commitment.purs
```diff
@@ -0,0 +1,59 @@
...
+import Data.Newtype (class Newtype, unwrap, wrap)
+import Data.Reflectable (class Reflectable)
+import Data.Vector (Vector)
+import Prelude ((<<<))
+import Snarky.Circuit.DSL.Monad (class CheckedType, check)
+import Snarky.Circuit.Types (class CircuitType, fieldsToValue, fieldsToVar, sizeInFields, valueToFields, varToFields)
+import Type.Proxy (Proxy(..))
+
+-- | IPA rounds in a Step proof (= log2 of Vesta SRS size = `Rounds.Step` = 16).
+type StepIPARounds = 16
```
martyall: Step and Wrap are names that only apply to pickles and should be kept here, why are they needed in kimchi?

packages/snarky-kimchi/src/Snarky/Backend/Kimchi/ProofCache.purs
```diff
@@ -48,15 +48,21 @@ import Foreign.Object as Object
...
+import Snarky.Backend.Kimchi.Proof (Proof, pallasProofFromSerdeJson, pallasProofToSerdeJson, vestaProofFromSerdeJson, vestaProofToSerdeJson)
 import Snarky.Backend.Kimchi.Types (VerifierIndex)
 import Snarky.Curves.Class (class PrimeField, toBigInt)
 import Snarky.Curves.Pallas as Pallas
 import Snarky.Curves.Vesta as Vesta
 
+-- | Step-side scalar field (= Vesta's scalar = Pallas's base field).
+-- | Inlined here rather than imported from `Pickles.Field` to keep
+-- | snarky-kimchi free of any pickles dependency.
+type StepField = Vesta.ScalarField
```
martyall: Step and Wrap names are only relevant for pickes, you should not use these aliases in kimchi

martyall (commented):
Are we sure that the proof cache is being used correctly for the tests? You would need to make sure the path is correct for whoever is invoking. I guess its fine sense its a absolute path

PR #147, opened 2026-06-14
Wasm web

martyall:


packages/example/src/Snarky/Example/Snark/Progress.purs
```diff
@@ -0,0 +1,82 @@
...
+-- |   * complete — the slot's proof is recorded
+-- |   * pending  — submitted but unproven (every leaf starts pending; a
+-- |                merge becomes pending the moment both children complete)
+-- |   * locked   — a merge whose work doesn't exist yet (children pending)
+-- |
+-- | Renderers live with their frontends: the terminal's log-update
+-- | display in the terminal package (ProgressDisplay), the browser's
+-- | SVG tree in the web package. Both consume the functions here, so
+-- | every frontend draws from the same status logic.
+module Snarky.Example.Snark.Progress
```
martyall, line 13: this module should be moved somewhere to example/terminal -- the to level lib is only used for genuinely shared code, and the browser app has it's own way of rendering this

packages/example/lib/src/Snarky/Example/Types/PublicKey.js
```diff
@@ -0,0 +1,55 @@
...
+//   base58( [0xCB, 0x01, 0x01] ++ x_LE(32) ++ [parity(y)] ++ checksum )
+//
+// where the compressed point is { x, is_odd = parity(y) } (the least
+// significant bit of y), x is the 32-byte little-endian field encoding
+// (exactly what `toHexLe` emits), the three prefix bytes are the
+// base58-check version (0xCB = non_zero_curve_point_compressed), the
+// non_zero_curve_point bin_prot version (0x01) and the compressed_poly
+// bin_prot version (0x01), and the checksum is the first 4 bytes of
+// sha256(sha256(version ++ payload)).
+import { sha256 } from "@noble/hashes/sha2.js";
```
martyall: is there not a standard, small base58 libary we can just use? which is usable in both node and web?

packages/example/lib/run-tests.mjs
```diff
@@ -0,0 +1,8 @@
+// Test launcher for the example lib. `spago test` cannot run under the
```
martyall: it's fine that the tests dont run via e backend optimizer, we don't need a special run script. only the application mains need this

packages/example/web/README.md
```diff
@@ -0,0 +1,41 @@
+# example-simulation-web
```
martyall: i think those whole readme could just be replaced with run instructions for the dev vs production version

packages/kimchi-napi/index.js
```diff
@@ -4,6 +4,24 @@
...
+// REQUIRED for multi-threaded proving.
+if (process.env.KIMCHI_BACKEND === "wasm") {
+  const k = require("./wasm/kimchi-napi.wasi.cjs");
+  // Full core count IN NODE: our napi calls are synchronous, so the
+  // main JS thread is blocked (asleep) during every Rust phase — it
+  // donates its core to the pool rather than contending with it.
+  // Measured: cores-1 costs ~8% prove wall here. The BROWSER is the
+  // opposite regime — the UI thread keeps running concurrently, so the
+  // browser init should use navigator.hardwareConcurrency - 1.
+  k.initThreadPool(require("os").cpus().length);
```
martyall, line 20: length or length - 1?

packages/kimchi-napi/package.json
```diff
@@ -4,12 +4,24 @@
   "description": "Local wrapper around the upstream kimchi-napi crate (built from the in-tree proof-systems submodule)",
   "main": "index.js",
   "scripts": {
-    "build": "bash build.sh"
+    "build": "bash build.sh",
+    "build:wasm": "cd ../../mina/src/lib/crypto/proof-systems && rustup target add wasm32-wasip1-threads && napi build --package kimchi-napi --package-json-path ../../../../../packages/kimchi-napi/package.json --target wasm32-wasip1-threads --profile release-wasm --platform -o ../../../../../packages/kimchi-napi/wasm && cd ../../../../../packages/kimchi-napi && node patch-wasi-browser.mjs && wasm-opt -O3 --enable-threads --enable-bulk-memory --enable-mutable-globals --enable-sign-ext --enable-nontrapping-float-to-int wasm/kimchi-napi.wasm32-wasi.wasm -o wasm/kimchi-napi.wasm32-wasi.wasm"
```
martyall: this looks wrong over overly complicated

packages/kimchi-napi/package.json
```diff
@@ -4,12 +4,24 @@
...
-}
+  "private": true,
+  "napi": {
+    "binaryName": "kimchi-napi",
+    "targets": [
+      "x86_64-unknown-linux-gnu",
+      "wasm32-wasip1-threads"
+    ],
+    "wasm": {
+      "initialMemory": 4000,
```
martyall, line 23: are these fixed by the browser?

packages/kimchi-napi/wasm-pool-config.mjs
```diff
@@ -0,0 +1,19 @@
...
+// The loader pre-spawns POOL_SIZE workers, shared between rayon and
+// emnapi's async-work pool. rayon must stay within
+// (POOL_SIZE - ASYNC_POOL_SIZE) threads: beyond that it spawns workers
+// on demand, which deadlocks — a nested worker's module script loads
+// via the (now blocked) parent thread's event loop.
+//
+// ASYNC_POOL_SIZE must match the `asyncWorkPoolSize` the napi CLI emits
+// into the loader (currently its default).
+
+export const POOL_SIZE = 16;
```
martyall: why are these baked in? There are 2 cases:
- web wasm: this feels wrong, you dont know the size of the thread pool
- node running wasm: these should just be build params or env vars with defauls

packages/kimchi-napi/patch-wasi-browser.mjs
```diff
@@ -0,0 +1,42 @@
+// Post-processes the napi-CLI-generated browser loader
```
martyall, line 1: How is it possible that napi-rs's solution for rayon requires patching

martyall (commented):
My biggest concern here is that there are two cases for wasm:
1. running in js host
2. running in web

And we have a bunch of ad-hoc modules, hard coded constatns, etc to deal with one or both. It just feels really unprinicpled. There are 2 phases: build and run and maybe all moules or scripes should be segmenetd as such

1. js host build (eg wasm build)
2. js host run
3. web host build (eg wasm build, vite packing
4. web host run
5. common

PR #155, opened 2026-06-20
ci: persist the Lagrange-basis cache across runs

martyall:
## What

Cache the kimchi **Lagrange bases** across CI runs via `actions/cache`, the same way we already cache the SRS `.srs` blobs.

The bases are a deterministic product of the fixed Pasta URS + the kimchi crypto — the same every run — but the memory-bound FFTs that build them cost minutes. Computing them once and injecting thereafter cuts that off the critical path of every CI run.

## How

- Point `SNARK_LAGRANGE_CACHE_DIR` at a repo-root `lagrange-cache/` (absolute, so spawned worker threads write the same dir) in the two jobs that actually build the cache:
  - **test-libs** — pickles `SharedSrs` → `fsCache`
  - **test-example** — the `run_simulation` terminal worker → `fsLagrangeCache`
  - `test-circuit-diffs` / schnorr build the SRS directly and never touch it, so they're left alone.
- Cache that dir under a **static** key. No need to hash the crypto version in: each cache filename already carries an SRS fingerprint (`blake2b` of the generators), so if the bytes ever change the fingerprint changes too — old entries miss, new ones recompute, never served wrong. The key is namespaced per-job (`github.job`) because test-libs and test-example warm *different* domain sets and would otherwise fight over one key (whoever saved first would freeze an incomplete set for the other).
- gitignore `/lagrange-cache/` (repo-root, like `/srs-cache/`); the `lagrange-cache` package under `packages/` stays tracked.

```yaml
- name: Cache Lagrange bases
  uses: actions/cache@v3
  with:
    path: lagrange-cache/
    key: ${{ runner.os }}-lagrange-${{ github.job }}
```

## Verification

Locally, with `SNARK_LAGRANGE_CACHE_DIR` set, a pickles run writes the fingerprint-keyed bases (`v1-{curve}-{blake2b}-d{N}`) to that exact dir (not tmpdir); the workflow YAML parses.

Builds on the `lagrange-cache` work already on `main`.


docs/srs-cache-lazy-plan.md
```diff
@@ -0,0 +1,232 @@
+# Lazy, persistent Lagrange-basis caching — threading the cache through compile + prove
```
martyall: delete the plan file

packages/lagrange-cache/src/Snarky/Lagrange/Cache/FS.purs
```diff
@@ -0,0 +1,76 @@
+-- | A filesystem-backed `LagrangeCache` for the Node runtime (tests, the
+-- | benchmark, the terminal app): one file per basis under a directory, so the
+-- | Lagrange bases survive across process runs. NODE ONLY — its FFI imports
+-- | `node:fs`; the browser uses `Snarky.Lagrange.Cache.Idb`.
+-- |
+-- | Synchronous (`Node.FS.Sync`), to feed the synchronous compile-time warm
+-- | without yielding (a yield would let a spec timeout preempt the cold FFT).
+module Snarky.Lagrange.Cache.FS
```
martyall, line 8: It feels like a lot of top level bindigns in this module should be demoted to where clauses

packages/lagrange-cache/src/Snarky/Lagrange/Cache/Hydrated.purs
```diff
@@ -0,0 +1,47 @@
...
+import Effect.Aff (Aff, launchAff_)
+import Effect.Class (liftEffect)
+import Effect.Ref as Ref
+import Snarky.Lagrange.Cache (LagrangeCache, keyString)
+
+-- | The async durable backing: bulk-load everything stored (to hydrate the map
+-- | up front, since a synchronous `get` cannot fetch on miss), and persist one
+-- | entry (run fire-and-forget on `put`). Keyed by `keyString`.
+type DurableStore =
+  { loadAll :: Aff (Map String Uint8Array)
```
martyall, line 32: String is a weak type, can we do better with a lagranch cache key type? This would also allow us to write the formatter from LagrangeCacheKey -> String

packages/lagrange-cache/src/Snarky/Lagrange/Cache/Idb.purs
```diff
@@ -0,0 +1,45 @@
...
+import Data.Array (zip)
+import Data.ArrayBuffer.Types (Uint8Array)
+import Data.Map as Map
+import Effect (Effect)
+import Effect.Aff (Aff)
+import Snarky.Lagrange.Cache (LagrangeCache)
+import Snarky.Lagrange.Cache.Hydrated (hydratedCache)
+
+-- | The shared same-origin database name used by every prover in the app.
+defaultDb :: String
```
martyall, line 28: should we newtype the db?

packages/lagrange-cache/src/Snarky/Lagrange/Cache.js
```diff
@@ -0,0 +1,5 @@
+// blakejs is CJS — default-import gives module.exports (the `{ blake2bHex, … }`
+// object); a named import does NOT work here. 16-byte (128-bit) digest, hex.
+import blakejs from "blakejs";
+
+export const srsFingerprint = (bytes) => blakejs.blake2bHex(bytes, undefined, 16);
```
martyall: Don't we already have a blake ps library?

packages/lagrange-cache/src/Snarky/Lagrange/Cache.purs
```diff
@@ -0,0 +1,148 @@
...
+  { curve: "pallas"
+  , srsToBytes: P.pallasSrsToBytes
+  , addBasis: P.pallasSrsAddLagrangeBasis
+  , basisToBytes: P.pallasSrsLagrangeBasisToBytes
+  , setBasisFromBytes: P.pallasSrsSetLagrangeBasisFromBytes
+  }
+
+-- | A 128-bit blake2b digest of `bytes`, hex. Synchronous (blakejs) so it runs
+-- | inside the compile-time warm, and browser-safe (unlike `crypto.subtle`).
+foreign import srsFingerprint :: Uint8Array -> String
```
martyall: we have a ps blake lib

packages/lagrange-cache/test/Test/Snarky/Lagrange/CacheSpec.purs
```diff
@@ -0,0 +1,48 @@
...
+import Test.Spec.Assertions (shouldEqual)
+
+spec :: SpecT Aff Unit Aff Unit
+spec = describe "Snarky.Lagrange.Cache (memoryCache + SRS fingerprint)" do
+  it "FFTs a domain once on a cold cache, then injects it onto a fresh same-URS SRS with no re-FFT" do
+    let
+      size = 8192
+      domain = 11
+      crs = V.vestaCrsCreate size
+    fftCount <- liftEffect (Ref.new 0)
```
martyall: consolidate usage of liftEffect

packages/lagrange-cache/test/Test/Snarky/Lagrange/FsCacheSpec.js
```diff
@@ -2,4 +2,4 @@ import { mkdtempSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
-export const mkTmpDir = () => mkdtempSync(join(tmpdir(), "srs-fscache-"));
+export const mkTmpDir = () => mkdtempSync(join(tmpdir(), "lagrange-fscache-"));
```
martyall: why do we need this foreign function?

packages/lagrange-cache/test/Test/Snarky/Lagrange/HydratedSpec.purs
```diff
@@ -0,0 +1,50 @@
...
+import Snarky.Lagrange.Cache.Hydrated (hydratedCache)
+import Test.Spec (SpecT, describe, it)
+import Test.Spec.Assertions (shouldEqual)
+
+foreign import mkBytes :: Int -> Uint8Array
+
+spec :: SpecT Aff Unit Aff Unit
+spec = describe "Snarky.Lagrange.Cache.Hydrated (hydrate/flush)" do
+  it "serves hydrated entries synchronously, and `put` is sync-visible + persists fire-and-forget" do
+    let
```
martyall, line 29: consolidate usage of liftEffect

PR #162, opened 2026-06-23
Formalize the EndoMul gate (GLV scalar mult) + factor shared EC lemmas

martyall:
**PR 4 of 4** — stacked on #161. Base retargets as the stack merges. The capstone: EndoMul, the endomorphism-optimized (GLV) variable-base scalar mul.

### Shared EC lemmas factored (first commit)
`secant_add`, `signed_target`, `some_eq_some` lifted out of VarBaseMul into the `Kimchi.Curve` oracle (reused by EndoMul). VarBaseMul's call sites resolve to them via parent-namespace lookup — no body changes.

### Gate — `Kimchi/Gate/EndoMul.lean`
VarBaseMul's `(P+Q)+P`, but each 2-bit window selects `Q ∈ {±T, ±φ(T)}` (GLV), `φ(x,y) = (endo·x, y)`.
- `Witness`/`Holds` — the **upstream-fixed** 12-constraint gate (incl. the distinct-point check `(xP−xR)·(xR−xS)·inv = 1`).
- `selectQ` (`±T`/`±φ(T)` by the bits), `block_sound` (one window's `(P+Q)+P`, via `secant_add` ×2), `distinctPoints`, `row_sound`, `row_int` (`∃ c₁ c₂, S = 4·P + c₁·T + c₂·φ(T)`).

> ⚠️ Adversarial verification (Aristotle) found the **pre-fix** 3-constraint block is *underconstrained* — `hc2`/`hc3` share a `(xP−xR)` factor and admit the spurious `R = −P` (counterexample on `y²=x³+1`: `P=(0,1),Q=(2,3),s=1,R=(0,−1)`). We model the **fixed** gate (o1-labs/proof-systems@64129ce4), so the non-degeneracy is *derived* (`distinctPoints`), not assumed. The repo's PureScript circuit stays pre-fix; the Lean models the sound gate.

### Circuit — `Kimchi/Circuit/EndoMul.lean`
- `chain_endo` (two-base fold) + `endoMul` (`∃ k₁ k₂, P_m = 4^m·P₀ + k₁·T + k₂·φ(T)`).
- `endoMul_scalar` — with the eigenvalue `φ(T) = [λ]·T` (a hypothesis), collapses to `P_m = 4^m·P₀ + k·T` — genuine scalar mult.
- `recoding_digit` — the `EndoMul ∘ EndoScalar` recoding correspondence: per 2-bit window, EndoScalar's `cPoly`/`dPoly` digits equal EndoMul's GLV window digit.

[...]

PR #160, opened 2026-06-23
Formalize VarBaseMul gate + circuit (variable-base scalar multiplication)

martyall:
**PR 2 of 2** — stacked on #159 (review that first; this PR's diff is only the VarBaseMul delta). Base will retarget to `main` once #159 merges.

Builds on the AddComplete gate to verify kimchi's **VarBaseMul** (signed-digit double-and-add) computes scalar multiplication, against Mathlib's group law.

### Gate — `Kimchi/Gate/VarBaseMul.lean`
Constraint model + reflection (`ok_iff`), then the per-gate soundness ladder:
`singleBit_sound` → `gate_scalarMul` (5 bits, point form) → **`gate_scalarMul_int`** — one satisfying gate computes `∃ c, P₅ = 32·P₀ + c·T ∧ (c:F) = 2·n' − 64·n − 31`, pinning the integer scalar to the gate's `n→n'` register.

### Circuit — `Kimchi/Circuit/VarBaseMul.lean`
- `chain_scalarMul` / `chain_register` — fold the per-gate relations over `m` gates.
- `scalarMul` — `P m = 32^m·P₀ + k·T`, `k` pinned to the register (`N 0 → N m`).
- `scalarMul_baseMul` — with `P 0 = a·T` the carry is absorbed: `P m = n·T`.
- `scalarMul_shifted` — at the real init (`[2]·T`, register `0`) the scalar is `2·N m + 2^(5m) + 1` = the pickles `Shifted_value.Type1.to_field`, **matching the reference value in proof-systems `varbasemul.rs`'s own test**.
- Type1 round-trip (`shiftType1` / `unshiftType1_shiftType1`) → **`scalarMul_caller`**: feed `shift(s)`, compute `[s]·T`. Type2 (`unshiftType2`, `scalarMul_type2`): split + the `if sOdd then h else h − T` correction.

The module docstring documents the **correspondence to the actual PureScript circuit** `Snarky.Circuit.Kimchi.VarBaseMul` — each hypothesis ↔ an emitted constraint, each theorem ↔ `scaleFast1`/`scaleFast2` (audit-level, by inspection).

All 20 headline theorems axiom-clean; build + style green in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #163, opened 2026-06-24
Lean: Pasta-cycle formalization — faithful two-field scalar gates (Level-1)

martyall:
Stacked on #162 (EndoMul). Makes the gates' **scalar** claims faithful to the two-field structure they actually live in.

## Why
The gate soundness proofs run over a single field `F`. For the **point** computations that's honest (group law + `[n]·P`, `n:ℤ`). But every **scalar** claim — VarBaseMul's `[s]·T`, EndoScalar's `a·λ+b`, `endoMul_toField` — silently identifies the coordinate field with the **scalar field** (group order). On Pasta these are *different* primes (Pallas coords `F_p`, order `q`; Vesta the mirror). The single-`F` statements are only correct under an unstated `Shifted_value` range condition — the very thing the cycle machinery exists to guarantee. This PR makes that explicit.

See `formal/docs/cycle-formalization.md` for the full scope.

## What's here (Level-1 — one curve, two fields)
All under `Kimchi.Cycle`, all axiom-clean (`propext, Classical.choice, Quot.sound`):

- **Phase 0 — `Cycle/Axioms.lean`**: `CMCurve`/`TwoCycle` bundle the facts Mathlib can't supply (curve order via Schoof, the CM eigenvalue `φ=[λ]`, reciprocity) as **structure fields**, so gate theorems parametrize over them and stay standard-axiom-clean. `zsmul_mod`: `[n]·P = [n mod order]·P`.
- **Phase 1 — `Cycle/VarBaseMul.lean`**: `varBaseMul_scalar` re-states VarBaseMul over a `CMCurve` with the multiplier a genuine `ℤ/order` element (`∀ s, n≡s → [s]·T`).
- **Phase 2 — `Cycle/Shifted.lean`**: `intCast_inj_of_sub_lt` (the cross-field coincidence: equal images + `|n−s|<p` ⟹ `n=s`) + `varBaseMul_faithful` — discharges Phase 1's conjunct for the `Shifted_value`-encoded scalar under an explicit range bound.
- **Phase 3 — `Cycle/EndoMul.lean`**: `endoMul_scalar_cm` / `endoMul_toField_cm` re-state EndoMul over a `CMCurve` with the **eigenvalue supplied by the curve** (`endoStep_eigen`), not hypothesized. `endoMul_faithful` closes the loop — `[EndoScalar.toField]·T` with both eigenvalue and scalar honest, modulo one explicit range hypothesis.

The key architectural property: the Pasta facts are *hypotheses* (`CMCurve` fields), so everything stays standard-axiom-clean. They only become real Lean `axiom`s when a curve is instantiated for Pasta — that's **Phase 4** (two-curve reciprocity + cross-cycle), deliberately out of scope here.

## Verification
`lake build Kimchi` green; 6 cycle headlines added to the CI sorry-gate (48 theorems, 0 `sorryAx`); `check-style.sh` clean.

### DIFF
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -95,6 +95,8 @@ jobs:
         #print axioms Kimchi.Circuit.EndoMul.recoding_digit
         #print axioms Kimchi.Circuit.EndoMul.sum_reindex
         #print axioms Kimchi.Circuit.EndoMul.recode_fold
+        #print axioms Kimchi.Circuit.EndoMul.row_digit
+        #print axioms Kimchi.Circuit.EndoMul.endoMul_ab
         #print axioms Kimchi.Circuit.EndoScalar.gate_toField
         #print axioms Kimchi.Circuit.EndoScalar.endoScalar_spec
         #print axioms Kimchi.Gate.VarBaseMul.ok_iff
--- a/.github/workflows/lean.yml
+++ b/.github/workflows/lean.yml
@@ -83,6 +83,9 @@ jobs:
         #print axioms Kimchi.Gate.EndoScalar.cPoly_table
         #print axioms Kimchi.Gate.EndoScalar.dPoly_table
         #print axioms Kimchi.Gate.EndoScalar.gate_endoScalar
+        #print axioms Kimchi.Gate.EndoScalar.gate_endoScalar_table
+        #print axioms Kimchi.Circuit.EndoScalar.gate_toField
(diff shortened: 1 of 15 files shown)


martyall:
Superseded by the interleaved stack: the cycle work now lives in #164 (foundation), #160 (VarBaseMul faithful), and #162 (EndoMul faithful + Pasta). Same content, organized so each gate carries its own faithful re-statement. Branch retained.

PR #167, opened 2026-06-25
formal: ingest the compiled constraint system via PS dump + verified plonkish checker 

kejace:
## What & why

Realizes `formal/Main.lean`'s original intent: **ingest a dumped `(circuit, witness)` and run a verified plonkish checker** over the exact gate list snarky elaborates — the object kimchi actually proves about — and then **prove end-to-end soundness of reconstructed multi-gate circuits over that same ingested model**.

This is the "Option B" direction: author circuits in PureScript, ingest the *compiled* constraint system into Lean, and prove about it — keeping snarky as the source of truth. It closes the faithfulness gap in the hand-transcribed gate proofs on two fronts: the Lean theorems are now about the precise dumped constraint system (drift becomes a CI failure), and the per-gate soundness results are *composed* into whole-circuit theorems.

## The ingested model + verified checker

- **`Kimchi/Circuit.lean`** — the faithful model (`Gate`/`Circuit`/`Witness`), a `rowHolds` dispatch over all 7 gate kinds, `gatesHold` + `copyHolds` (the extensional permutation/copy-constraint relation) + `Satisfies`, an executable `check`, and the **`check_iff`** reflection theorem (proven symbolically — never `decide`, so it holds at the 255-bit Pasta field).
- **`Kimchi/Gate/{VarBaseMul,EndoMul,EndoScalar,Poseidon}.lean`, `Gate/AddComplete.lean`, `Gate/Generic.lean`** — the real constraint polynomials transcribed from kimchi `proof-systems` (21 / 11 / 11 / 15 constraints), replacing the earlier `True` stubs, each with `Holds`/`ok`/`ok_iff` and algebraic soundness against Mathlib's group law.
- **`Kimchi/Json.lean` + `Main.lean`** — LE-hex parsing, `FromJson`, column→row transpose, `loadAndCheck`.

## Circuit composition proofs (`Kimchi/Circuits/*Step.lean`)

Reconstruct the *actual* dumped circuit and prove **∀-witness soundness over `Satisfies`**, deriving the wiring (copy constraints) and each row's gate identity (via a `checker_holds_iff` bridge) from the circuit itself. Only curve-level nonsingularity / non-degeneracy stay as hypotheses (as `AddCompleteStep` keeps input nonsingularity a hypothesis). All five gates:

- **`AddCompleteStep`** — a `CompleteAdd` row certifies a genuine EC addition of the public inputs.
- **`VarBaseMulStep`** — an `m`-gate `VarBaseMul`/`Zero` chain computes `[n]·T`; threading derived from `copyHolds`.
- **`EndoMulStep`** — the GLV endo-mul chain computes `[s]·T`. The dumped gate omits the distinct-point (`inv`) constraint, so the bridge is an iff *up to* `(xP−xR)(xR−xS) ≠ 0` (a curve precondition), with `ofRows` choosing `inv := ((xP−xR)(xR−xS))⁻¹`.
- **`EndoScalarStep`** — the endo-scalar decomposition computes `a·λ + b`. The dumped `a`/`b` constraints are `6·`-scaled, so the bridge needs `2, 3 ≠ 0` (the division isolated in `c6_eq`/`d6_eq`, peeled by `mul_left_cancel₀`).
- **`PoseidonStep`** — the reconstructed chain computes the `5m`-round Poseidon permutation of the initial state (no external Mathlib spec — the gate *defines* the permutation).

### Pasta specializations
Routing `Satisfies` through the deployed Pasta roots discharges the curve non-degeneracy / char / order side conditions internally: `varBaseMul_circuit_scaleFast1/2`, `pallas_endoMul_circuit`, `EndoScalar.{pallas,vesta}_circuit_sound`.

### Scalar-field lift (`exists_canonical_scalar`)
The integer output scalar `n` is reduced to its canonical residue in `[0, order)` — an element of the 2-cycle sister curve's base field (the cross-field value Pickles' Wrap verifier consumes), via the existing `zsmul_mod`/`order_smul`, no new axioms.

### Init grounding (fully end-to-end)
The chain's initial state, previously a hypothesis, is now *derived* from setup rows for both techniques:
- **`EndoScalar.esCircuitGrounded_sound`** — constant-pinning: `Generic` setup rows pin `(2,2,0)` into row 0.
- **`VarBaseMul.vbmCircuitGrounded_scaleFast1`** — gate dataflow: a `CompleteAdd` doubling row produces `[2]·T`, wired into gate 0's accumulator; the four wiring equalities are read off `copyHolds`.

### Heterogeneous combinations (`Combinations.lean`)
For every unordered pair of the five custom kinds, a two-block example circuit whose `Satisfies` yields *both* gates' algebraic facts — demonstrating the ingested model composes gate kinds.

### Shared combinators & statement strengthening
A review pass consolidated the reconstruction machinery: **`Circuit.append`** + `gateAt_append_left/right` + **`Satisfies.of_append`** (the "proven block plus extra rows" pattern, previously hand-rolled per circuit) — `vbmCircuitGrounded` and `esCircuitGrounded` are now one-line appends; `getD_ofFn_lt` and `Point.some_congr` moved to their single shared homes. And a strengthening: **`Point.add_self_ne_zero`** (no affine 2-torsion off the x-axis) lets the init-doubling theorems *derive* the `CompleteAdd` `inf = 0` flag from the full `AddComplete.sound` disjunction instead of assuming it — the `hinf` hypotheses are gone.

## The axiom story

Builds clean, **no `sorry`**. The trusted base is now just **four postulates**: the two Pasta Hasse bounds (`{pallas,vesta}_hasse`, from which the group orders are *derived* via CompElliptic) and the two CM eigenvalue relations (`{pallas,vesta}_eigen`, that `φ(x,y)=(β·x,y)` acts as `[λ]`). The endomorphism coefficients `{pallas,vesta}_endo` were **turned from opaque axioms into concrete `def`s** (the real `endoBase` numerals) and machine-checked to be primitive cube roots of unity (`{pallas,vesta}_endo_cube`, by kernel `decide`) — removing two axioms and letting the EndoMul specialization drop its `endo = β` hypothesis. `scripts/check_axioms.sh` gates **79 headline roots** to reduce to exactly this allowed set (plus the standard logical axioms + `Lean.ofReduceBool`) and fails on any `sorry`.

## Empirical validation

- **Gate transcriptions:** each committed fixture under `formal/fixtures/` (`add_complete`, `varbasemul`, `endomul`, `endoscalar`, `poseidon`) is a *real solver-produced witness*, and the Lean `check` returns `true` on it (tamper any cell → `false`) — so the constraints match kimchi's exactly, including the hardcoded constants.
- **Reconstructions:** `CheckReconstruction.lean` slices each multi-gate fixture to its gate chain and confirms the hand-written reconstruction (`vbmCircuit`/`emCircuit`/`esCircuit`/`posCircuit`) accepts the real witness and rejects a tamper — machine-verifying the column layout / copy wiring the composition proofs assume.

## PureScript

`exportCircuitWitness` emits the combined `{circuit, witness}` JSON; a self-contained `DumpAddComplete` test compiles → solves → dumps the real circuits.

## CI

The Lean workflow runs, over the committed fixtures: `make lean-check-fixtures` (the checker on every fixture), `make lean-check-reconstruction` (the reconstructions accept the real chains, reject tampers), and the axiom gate (`scripts/check_axioms.sh`, 79 roots). All green.

## Scope / trusted base / known seam

- SNARK soundness (that kimchi *enforces* the constraints) stays axiomatic; the `Spec` is a human statement of intent; the trusted base is the four Pasta postulates above.
- **The one honest seam:** the composition proofs are stated over hand-written reconstructions of the dumped circuits and validated against real dumps *empirically* (`CheckReconstruction`), not by a proof that the dumper emits exactly that circuit. `EndoScalar.esCircuitGrounded_sound`'s setup rows likewise idealize the dump's double-`Generic` packing. Both are documented in-code; closing the seam fully means reasoning about the dump's global permutation cycles and is out of scope here.

## Follow-ups

`formal/docs/composition-followups.md` (the #4/#6 designs, now implemented) and `formal/docs/breadth-plan.md` (extending the checker to the remaining kimchi gates — RangeCheck/ForeignField/Xor/Rot/Lookup — with the PureScript-gadget dependency and the lookup-entanglement fork).


formal/Kimchi/Checker/EndoMul.lean
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Kimchi.Checker.EndoMul
+
+open Kimchi.Gate (Row)
+
+variable {F : Type*}
+
+/-- The base-field endomorphism coefficient `endoBase` for the Pallas curve (`= Fp`),
+    taken from the PureScript backend (`PastaCurve.js` `pallasEndoBase`). -/
+def endo [CommRing F] : F :=
```
martyall, line 30: You should store these in a central location. As of right now, the exact value hasn't been important so you can overwrite it
https://github.com/l-adic/snarky/blob/main/formal/Kimchi/Pasta.lean#L88

formal/Kimchi/Checker/EndoMul.lean
```diff
@@ -0,0 +1,79 @@
...
+  20444556541222657078399132219657928148671392403212669005631716460534733845831
+
+def holds [CommRing F] (curr next : Row F) : Prop :=
```
martyall, line 33: all of these are already defined in Kimchi/Gate/EndoMul, 

https://github.com/l-adic/snarky/blob/main/formal/Kimchi/Gate/EndoMul.lean#L142

the only difference is `Row` vs `Witness`. But the `witness` is just 2 rows, so you should unpack it and use the defining equations from there

formal/Kimchi/Checker/EndoMul.lean
```diff
@@ -0,0 +1,79 @@
...
+  ∧ (2 * w 4 - w 9 ^ 2 + xq1) * ((w 4 - w 7) * w 9 + w 8 + w 5) - (w 4 - w 7) * (2 * w 5) = 0
+  ∧ (w 8 + w 5) ^ 2 - (w 4 - w 7) ^ 2 * (w 9 ^ 2 - xq1 + w 7) = 0
+  -- second endo-addition (slope s3, output point S in next row)
+  ∧ (xq2 - w 7) * w 10 - (yq2 - w 8) = 0
+  ∧ (2 * w 7 - w 10 ^ 2 + xq2) * ((w 7 - wn 4) * w 10 + wn 5 + w 8) - (w 7 - wn 4) * (2 * w 8) = 0
+  ∧ (wn 5 + w 8) ^ 2 - (w 7 - wn 4) ^ 2 * (w 10 ^ 2 - xq2 + wn 4) = 0
+  -- scalar accumulation
+  ∧ 16 * w 6 + 8 * w 11 + 4 * w 12 + 2 * w 13 + w 14 - wn 6 = 0
+
+def ok [CommRing F] [DecidableEq F] (curr next : Row F) : Bool :=
```
martyall, line 55: also dfined in Endomul Gate module (or easily written from https://github.com/l-adic/snarky/blob/main/formal/Kimchi/Gate/EndoMul.lean#L115)

formal/Kimchi/Gate/Poseidon.lean
```diff
@@ -0,0 +1,78 @@
+import Kimchi.Checker.Row
```
martyall, line 1: There should be a single PR to only add Poseidon gate with no other changes.

formal/docs/breadth-plan.md
```diff
@@ -0,0 +1,114 @@
+# Breadth plan: widening the ingested checker to the rest of the kimchi gate set
```
martyall, line 1: dont commit plan files

formal/docs/composition-followups.md
```diff
@@ -0,0 +1,120 @@
+# Composition-proof follow-ups: plans for #4 and #6
```
martyall, line 1: dont commit plan files

packages/pickles-circuit-diffs/src/Pickles/CircuitDiffs/Circuit.purs
```diff
@@ -93,6 +95,43 @@ comparable c =
...
   }
 
+--------------------------------------------------------------------------------
+-- Combined {circuit, witness} export for the Lean verified checker
+
+-- | The combined dump ingested by `formal/Kimchi/Json.lean`: gates with their
+-- | little-endian-hex coefficients, the column-major (15 × n) witness, and the public
+-- | inputs. Field elements `f` serialize as LE hex through their `WriteForeign` instance,
+-- | matching the OCaml circuit fixtures.
+type CircuitWitnessExport f =
```
martyall, line 105: `pickles-circuit-diffs` package was made to test constraint systems vs ocaml pickles fixtures

if you want to do witness generation, you could likely re-use what's already there (which is a lot of circuits) and add just some flag / config option which will do witness generation in addition to what it's already doing. Then you don't have to write a separate witness dump for everything

martyall (changes requested):
We need to get PRs down to like max +500/-500 in order to be able to track what's going on. This could mean taking one gate at a time rather than all at once.

I left some general comments for EndoMul about duplication which apply all over here.

My recommendation would be to start us off by a series of independent PRs
1. submitting a single PR for poseidon gate that follows all existing patterns.
2. Change the `pickles-circuit-diffs` package to work with some config option that when true will do witness generation and write it to a file in a way which is coherent ith what it already does. You can then run this to make sure it does what you think it should
3. Set up the general parse machinery and apply it for a  single gate, eg add complete

kejace:
Closed in favor of #187 (and subsequent PRs)

PR #183, opened 2026-07-02
formal: define the double generic gate in Gate/Generic.lean; Quotient consumes it

martyall:
Follow-up to #182, fixing a layering mistake: `Quotient/Generic.lean` defined its own row predicate (`GenericHolds`) instead of working out of the `Kimchi/Gate` modules. Gate semantics belong in `Gate/`; higher layers lift them.

- **`Gate/Generic.lean`** gains the double generic row, defined as **literally two existing `Generic` gates** packed on one row (`DoubleGeneric.fst/snd` from the coefficient cells, `registers` as the row's first six cells, `Holds` as the conjunction of the two `Generic.holds`), plus `holds_iff` unfolding to the raw `generic.rs` l.245–250 cell equations.
- **`Quotient/Generic.lean`**: quotient-local `GenericHolds` deleted; `genericRows_iff_dvd` restated against `Gate.DoubleGeneric.Holds`. The file now owns only the polynomial side (constraint polynomials over column interpolants + the divisibility bridge).

No statement content changes beyond the predicate's home; roots unchanged (38), axiom closure unchanged (`genericRows_iff_dvd` still `[propext, Classical.choice, Quot.sound]`). Full build green (11442 jobs), style clean.

### DIFF
--- a/formal/Kimchi/Gate/Generic.lean
+++ b/formal/Kimchi/Gate/Generic.lean
@@ -69,6 +69,45 @@ theorem satisfies_iff [DecidableEq F] (gs : List (Generic F)) (a : Assignment F)
     satisfies gs a = true ↔ Satisfies gs a := by
   simp [satisfies, Satisfies, List.all_eq_true, Generic.ok_iff]
 
+/-! ## The double generic gate row
+
+kimchi's generic gate row (`generic.rs`) packs **two** generic gates: the first on registers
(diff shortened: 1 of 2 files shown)


martyall:
Superseded by the gate-constraints refactor PR (the Generic gate was redesigned as the double generic row directly, and the constraint-expression single-sourcing now covers all gates).

PR #186, opened 2026-07-04
formal: kimchi quotient soundness engine + gate lifts

martyall:
The quotient argument's soundness layer, built on the constraints-as-data foundation of #184. With this PR the chain **gate semantics → constraint polynomials → Z_H divisibility → random-evaluation checks** is formalized end-to-end for Generic, AddComplete, VarBaseMul, and EndoMul.

Everything is commitment-free pure polynomial algebra over `[Field F]` with a primitive-root hypothesis — **no trust boundary anywhere in this layer**: no axioms, no FS-style hypotheses; distinct-challenge premises are ordinary theorem hypotheses (their Fiat–Shamir packaging is a later capstone's job).

## The soundness engine

- **`Pinning.lean`** — `identity_of_evals` / `zH_dvd_of_evals`: the quotient identity `C(ζ) = t(ζ)·Z_H(ζ)` at more-than-degree-many distinct points holds as polynomials (abstract degree bound; too-many-roots engine).
- **`Aggregate.lean`** — `aggregate α E = Σ α^c • E c` (kimchi's consecutive alpha powers) and `dvd_separation`: the aggregate divisible at enough distinct α's forces each constraint divisible.
- **`Shifted.lean`** — `shift ω W = W.comp (C ω·X)` with the cyclic next-row eval bridge (the two-row-gate mechanism).

## The lift engine + gate lifts

- **`Lift.lean`** — `rows_iff_dvd_of` and the selector-gated `rowsSel_iff_dvd`, stated **once** for every gate, parametrized by an eval bridge: `constraints` being a natural transformation (`constraints_map`, #184) makes the per-gate bridge a `simp`.
- **`AddComplete/VarBaseMul/EndoMul.lean`** — each file is exactly a cell map (the gate's witness layout) plus that bridge. The constraint polynomials are `Gate.<G>.constraints` read over `F[X]` — **no gate formula is restated anywhere in this layer.**
- Faithfulness note: EndoMul's `inv` cell (current-row column 2) is verified against the upstream fix o1-labs/proof-systems@64129ce4 (11 → 12 constraints). The vendored `endosclmul.rs` is the pre-fix gate; our Gate module deliberately models the fixed one (documented in the file).

## Headline

- **`Soundness.lean`** — per gate: evaluations at enough distinct (ζ, α) ⟹ every selector-active row satisfies `Holds` (`<G>.soundness`), and via the landed `Gate.<G>.sound`, the EC meaning (`<G>.soundness_point`) — cited, not reproven.

## Verification

`check_axioms` roots 38 → 48; every new root closes at `[propext, Classical.choice, Quot.sound]`. Full build green (11450 jobs), style clean (39 files).

Deferred (documented in module headers): EndoScalar's per-row reformulation + coefficient genericity, permutation, lookup, linearization/`ft`, concrete degree bounds, chunking.

### DIFF
--- a/formal/Kimchi/Quotient/AddComplete.lean
+++ b/formal/Kimchi/Quotient/AddComplete.lean
@@ -1,0 +1,112 @@
+import Kimchi.Quotient.Lift
+import Kimchi.Gate.AddComplete
+
+/-!
+# Quotient lift of the CompleteAdd gate
+
+Archon-original polynomial-algebra lift of kimchi's CompleteAdd gate, built on the generic
+lift engine (`Kimchi.Quotient.Lift`) and the domain substrate (`Kimchi.Quotient.Domain`).
+
+CompleteAdd is a **single-row** gate, so its cell map reads only the current row. The gate's
+field-level constraint model (`Kimchi.Gate.AddComplete.constraints` / `Holds`) is READ-ONLY
+and reused verbatim: no constraint formula is restated here — the lift is purely the
+naturality of `constraints` under evaluation at the domain nodes.
+
+## Contents
+
+* `cellMap` — assemble a `Gate.AddComplete.Witness R` from a row `cur : Fin 15 → R`.
+* `rowWitness` / `polyWitness` — the row-values and column-interpolant witnesses, both via
+  the same `cellMap`.
+* `bridge` — evaluating the gate's constraint polynomials at `ω^i` reproduces the gate's
+  constraints at row `i` (pure naturality).
+* `rows_iff_dvd` / `rowsSel_iff_dvd` — the two divisibility corollaries, immediate instances
+  of the engine theorems.
+
+Source of truth: `blueprint/src/chapters/Kimchi_Quotient_AddComplete.tex`.
+-/
+
+namespace Kimchi.Quotient.AddComplete
+
+open Polynomial
+
+variable {F : Type*} [Field F] {n : ℕ} {ω : F}
+
+/-! ## Column layout and the cell map
+
+A CompleteAdd row occupies the 11 witness columns `0`–`10` of the size-`15` circuit table;
+columns `11`–`14` are unused. Layout (kimchi `complete_add.rs`, module-doc table):
+
+```
+|  0 |  1 |  2 |  3 |  4 |  5 |  6  |    7   | 8 |   9   |    10   |
+| x1 | y1 | x2 | y2 | x3 | y3 | inf | same_x | s | inf_z | x21_inv |
+```
+-/
+
+/-- **CompleteAdd cell map.** Assemble a `Gate.AddComplete.Witness R` by reading the circuit
+columns of a row `cur : Fin 15 → R`. -/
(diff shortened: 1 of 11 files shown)


formal/Kimchi/Quotient/AddComplete.lean
```diff
@@ -0,0 +1,112 @@
+import Kimchi.Quotient.Lift
```
martyall, line 1: this exact pattern is repeated across every gate. Its elegantly stated but repetative, is there no typeclass like primitive that states it for a generic gate that could just be applied to all of our gate?

formal/Kimchi/Quotient/Aggregate.lean
```diff
@@ -0,0 +1,95 @@
+import Kimchi.Quotient.Domain
+
+/-!
+# α-aggregation and constraint separation
+
+This file is **Archon-original** polynomial-algebra infrastructure for kimchi's quotient
```
martyall: dont mention archon

formal/Kimchi/Quotient/Soundness.lean
```diff
@@ -0,0 +1,336 @@
+import Kimchi.Quotient.Pinning
+import Kimchi.Quotient.Aggregate
+import Kimchi.Quotient.AddComplete
+import Kimchi.Quotient.VarBaseMul
+import Kimchi.Quotient.EndoMul
+
+/-!
+# The quotient-argument soundness
+
+Archon-original. The headline of the commitment-free quotient argument: it composes the
```
martyall: dont mention archon

martyall (commented):
what about the quotient argument for EndoScalar?

Do we really need the soundness point arguments for every gate? The reason I had introduced those was to give semantic meaning to the gates, but I doubt the play any real role in the soundness of the protocol

martyall:
Review applied in the latest commit:

- **`soundness_point` dropped** (all three gates) — agreed: protocol soundness composes at the `Holds` level; the EC meaning belongs where it's consumed, via `Gate.<G>.sound`. The Holds-level `<G>.soundness` headlines replace them in the axiom-gate roots (still 48, all closing at the standard three).
- **Process prose scrubbed** — the two flagged "archon" mentions plus five more of the same across the module docstrings, and the "later wave" stage-jargon in proof docstrings.

Deferred to the next brick (agree-but-needs-discussion items from the review):

- **EndoScalar quotient argument** — needs the per-row reformulation (the gate's `Witness` models a multi-row fold; the quotient needs the actual `endomul_scalar.rs` single-row layout) and a genericity story for `cPoly`'s field-constant coefficients.
- **The per-gate repetition** — a `GateLift F` structure (witness functor + `constraints` + naturality + cell map) with `rows_iff_dvd`/`rowsSel_iff_dvd` proven once over it. The unlock for making all four gates fit one primitive: state naturality over `F`-**algebra** homs instead of ring homs — then EndoMul's `endo` and EndoScalar's `2/3` are both `algebraMap` images and `eval (ω^i) : F[X] →ₐ[F] F` still instantiates it. Touches the same files as the EndoScalar brick, so the two go together.

Build green (11450 jobs), 48/48 roots, style clean.

PR #188, opened 2026-07-04
[03/13] formal: constraint-system ingestion + verified checker (AddComplete)

kejace:
Third of the series decomposing #167 (see [01/13] … [13/13]). The general **constraint-system ingestion + verified checker**, applied to a **single gate (AddComplete)** — the review's step 3. Builds on `Gate/AddComplete` (already on main); independent of [01/13]/[02/13] at the code level.

## What

- **`Kimchi/Circuit.lean`** — the faithful Lean model of the dumped gate list: `Gate` / `Circuit` / `Witness`, a `rowHolds` dispatch, `gatesHold` + `copyHolds` (the extensional copy-constraint relation) + `Satisfies`, an executable `check`, and the **`check_iff`** reflection theorem. Proven **symbolically — never `decide`** — so it holds at the 255-bit Pasta field. `rowHolds` is sliced to **Generic / Zero / CompleteAdd** here; the four custom gates are `True`-stubbed and filled one PR at a time (`[04/13]`…`[07/13]`).
- **CompleteAdd is derived from the gate, not re-transcribed.** `rowHolds`'s completeAdd case is `AddComplete.Holds (AddComplete.ofRow curr)`, where `ofRow` (added to `Gate/AddComplete.lean`) unpacks the 15-column row into the gate's named `Witness` — so the entire `Gate/AddComplete` soundness suite applies verbatim. This is the de-duplication pattern from the review, clean from the first checker gate. (The re-transcription the review flagged only affects the *separate* VarBaseMul/EndoMul/EndoScalar checkers, which come in `[04/13]`–`[06/13]` and will follow the same derive-from-gate approach.)
- **`Kimchi/Json.lean` + `Main.lean`** — LE-hex parsing, `FromJson`, column→row transpose, `loadAndCheck`; `Main` checks every `fixtures/*.json` (or explicit paths), exiting non-zero on any `false`. **`Kimchi/Field/Pasta.lean`** supplies the concrete `Fp`/`Fq`.
- **`fixtures/add_complete_step.json`** — a real solver-produced witness.
- **`make lean-check-fixtures`** — runs the checker over every fixture in CI.

## Checks

- `check fixtures/add_complete_step.json = true`; tampering any cell → `false` (verified via the `make lean-check-fixtures` target, exit 0)
- Full `Kimchi` target builds green; `check_iff` is axiom-clean (`[propext, Quot.sound]`)
- `formal/scripts/check-style.sh` clean

## Relation to [02/13]

The witness-dump helper in `[02/13]` (`WitnessDump.dumpWitnessTo`) emits exactly this `{ publicInputSize, gates, witness, publicInputs }` shape — it is the same code that produced the fixture here — so harness-generated fixtures ingest directly. The two were developed in tandem.


formal/Kimchi/Field/Pasta.lean
```diff
@@ -0,0 +1,58 @@
...
+/-- Pallas base field characteristic (Vesta scalar field). -/
+def pastaP : ℕ :=
+  0x40000000000000000000000000000000224698fc094cf91b992d30ed00000001
+
+/-- Vesta base field characteristic (Pallas scalar field). -/
+def pastaQ : ℕ :=
+  0x40000000000000000000000000000000224698fc0994a8dd8c46eb2100000001
+
+/-- **AXIOM.** `pastaP` is prime. (Published Pasta-cycle fact; see module docstring.) -/
+axiom pastaP_prime : Nat.Prime pastaP
```
kejace, line 46: We can either wait for CompElliptic for this, or roll or own. Either way I have done it before and it's not a blocker

formal/Kimchi.lean
```diff
@@ -9,6 +9,15 @@ import Kimchi.Gate.AddComplete
 import Kimchi.Gate.VarBaseMul
 import Kimchi.Gate.EndoScalar
 import Kimchi.Gate.EndoMul
+
+-- The ingested constraint-system model + verified plonkish checker ([03/13]): the faithful
```
kejace: remove reference to PR here

kejace:
Folded into #189 per review — the checker now ships together with the witness-generation output and the AddComplete example end-to-end. Remaining gates follow in the next PR.

PR #189, opened 2026-07-04
[02/13] witness output from the diff harness + Lean verified checker (AddComplete)

kejace:
Continues the decomposition of #167 — this PR carries **the witness-generation output _and_ the Lean ingestion + verified checker, with AddComplete as the worked gate example end-to-end** (the former #188 is folded in here, per review). The remaining gates come in the next PR.

## PureScript: witness output from the diff harness

- **`Pickles.CircuitDiffs.Types.WitnessExport`** — a self-contained circuit-satisfaction instance `{ publicInputSize, gates, witness, publicInputs }`, all field elements LE-hex (`toHexLe`). **`CircuitComparison` gains `witness :: Maybe WitnessExport`** — the comparison JSON in `circuits/results/` carries the solved witness; older results JSONs still parse, the visualizer is a read-only consumer.
- **`WitnessDump.buildWitnessExport`** — run the solver on a concrete input, package the export.
- **`exactMatchWitnessEff @a @b name circuit input`** — `exactMatchEff` with one shared compilation feeding both the OCaml diff and the embedded `witness` field. **Only `add_complete_step_circuit` uses it in this PR** — the one gate the checker below constrains. The other four gate registrations are untouched and switch over with their checkers in the next PR.

## Lean: ingestion + verified checker (AddComplete)

- **`Kimchi/Circuit.lean`** — the faithful model of the dumped gate list (`Gate`/`Circuit`/`Witness`, `rowHolds` dispatch, `gatesHold` + `copyHolds` + `Satisfies`), an executable `check`, and the **`check_iff`** reflection theorem — proven symbolically (never `decide`), so it holds at the 255-bit Pasta field; axiom-clean (`[propext, Quot.sound]`). `rowHolds` constrains **Generic / Zero / CompleteAdd**; the custom gates are `True`-stubbed and filled per-gate in follow-ups.
- **CompleteAdd is derived from the gate, not re-transcribed**: the completeAdd case is `AddComplete.Holds (AddComplete.ofRow curr)` — `ofRow` unpacks the 15-column row into the gate's named `Witness`, so the whole `Gate/AddComplete` soundness suite applies verbatim (the de-duplication pattern from the review).
- **No committed fixture — `circuits/results/` is the single source.** `Json.loadAndCheck?` reads either a raw dump or a comparison JSON's `witness` field; `Main.lean`'s no-args mode scans the results directory, checks every witness-carrying comparison, and fails if any check fails or none exist. The check always runs against a freshly-generated witness, so it cannot go stale. `make lean-check-witnesses` wraps it (run the harness first).

## Verified live, end-to-end

```
spago test -p pickles-circuit-diffs -- --example add_complete_step_circuit
  ✓ add_complete_step_circuit matches OCaml        (witness generation inside the registration)
lake env lean --run Main.lean
  check .../circuits/results/add_complete_step_circuit.json = true
  checked 1 witness-carrying result(s)             (tampering a witness cell → false)
```

PS: harness + types type-check clean, purs-tidy formatted. Lean: full build green on current main, style + axiom gates pass.

### DIFF
--- a/packages/pickles-circuit-diffs-types/src/Pickles/CircuitDiffs/Types.purs
+++ b/packages/pickles-circuit-diffs-types/src/Pickles/CircuitDiffs/Types.purs
@@ -2,6 +2,7 @@ module Pickles.CircuitDiffs.Types
   ( ComparableGate
   , ComparableCircuit
   , CircuitComparison
+  , WitnessExport
   ) where
 
 import Data.Maybe (Maybe)
@@ -14,10 +15,18 @@ type ComparableGate =
   , context :: Array String
   }
 
+-- | A solved witness for the circuit it hangs off of: the 15 register columns
+-- | (column-major) and the public-input values, all field elements as LE-hex strings.
+type WitnessExport =
+  { witness :: Array (Array String)
+  , publicInputs :: Array String
+  }
+
 type ComparableCircuit =
   { publicInputSize :: Int
   , gates :: Array ComparableGate
   , cachedConstants :: Array { variable :: Int, varType :: String, value :: String }
+  , witness :: Maybe WitnessExport
   }
 
 type CircuitComparison =
--- a/packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
+++ b/packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
@@ -5,8 +5,10 @@ import Prelude
 import Control.Monad.Rec.Class (Step(..), tailRecM)
 import Data.Array as Array
 import Data.Either (Either(..))
+import Data.Int as Int
 import Data.Int.Bits as Bits
-import Data.Maybe (Maybe(..))
+import Data.Maybe (Maybe(..), fromMaybe)
+import Data.Monoid (power)
 import Data.Newtype (un)
 import Data.Tuple (Tuple(..))
 import Data.Vector (Vector, (:<))
@@ -14,13 +16,15 @@ import Data.Vector as Vector
 import Effect (Effect)
 import Effect.Aff (Aff)
 import Effect.Class (liftEffect)
+import Effect.Console as Console
 import Effect.Exception (throw)
 import Node.Buffer as Buffer
 import Node.Encoding (Encoding(..))
 import Node.FS.Perms (all, mkPerms)
 import Node.FS.Sync as FS
 import Node.Process as Process
-import Pickles.CircuitDiffs.Circuit (Circuit, ComparableCircuit, comparable, fromCompiledCircuit, parseOcamlFixtures)
+import Partial.Unsafe (unsafeCrashWith)
+import Pickles.CircuitDiffs.Circuit (Circuit, ComparableCircuit, comparable, fromCompiledCircuit, fromGateData, gateDataOf, parseOcamlFixtures)
 import Pickles.CircuitDiffs.PureScript.BCorrect (compileBCorrect)
 import Pickles.CircuitDiffs.PureScript.BulletReduce (compileBulletReduce)
 import Pickles.CircuitDiffs.PureScript.BulletReduceOne (compileBulletReduceOne)
@@ -68,8 +72,9 @@ import Pickles.CircuitDiffs.PureScript.WrapVerify (compileWrapVerify)
 import Pickles.CircuitDiffs.PureScript.WrapVerifyN2 (compileWrapVerifyN2)
 import Pickles.CircuitDiffs.PureScript.Xhat (compileXhat)
 import Pickles.CircuitDiffs.PureScript.XhatStep (compileXhatStep)
-import Pickles.CircuitDiffs.Types (CircuitComparison)
+import Pickles.CircuitDiffs.Types (CircuitComparison, WitnessExport)
 import Pickles.PublicInputCommit (mkConstLagrangeBaseLookup)
+import Random.LCG (mkSeed)
 import Safe.Coerce (coerce)
 import Simple.JSON (writeJSON)
 import Snarky.Backend.Advice (noAdvice)
@@ -82,7 +87,7 @@ import Snarky.Backend.Kimchi.Impl.Vesta (vestaCrsCreate)
 import Snarky.Backend.Kimchi.Proof (createProof)
 import Snarky.Backend.Kimchi.Types (CRS)
 import Snarky.Circuit.CVar (add_) as CVar
-import Snarky.Circuit.DSL (class BasicSystem, BoolVar, F(..), FVar, SizedF, addConstraint, all_, and_, any_, assertEqual_, assertNonZero_, assertNotEqual_, assertSquare_, assert_, const_, div_, equals_, exists, if_, inv_, mul_, or_, pow_, unpack_, xor_)
+import Snarky.Circuit.DSL (class BasicSystem, class CheckedType, class CircuitType, BoolVar, F(..), FVar, SizedF, addConstraint, all_, and_, any_, assertEqual_, assertNonZero_, assertNotEqual_, assertSquare_, assert_, const_, div_, equals_, exists, if_, inv_, mul_, or_, pow_, unpack_, xor_)
 import Snarky.Circuit.DSL.Monad (Snarky)
 import Snarky.Circuit.Kimchi.AddComplete (Finiteness(..), addFast)
 import Snarky.Circuit.Kimchi.EndoMul (endo)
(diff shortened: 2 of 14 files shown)


packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/WitnessDump.purs
```diff
@@ -0,0 +1,127 @@
...
+-- | `{ publicInputSize, gates, witness, publicInputs }` shape the Lean ingestion reads.
+dumpWitnessTo
+  :: forall @f g a b
+   . CircuitGateConstructor f g
+  => PrimeField f
+  => WriteForeign f
+  => String
+  -> CircuitBuilderState (KimchiGate f) (AuxState f)
+  -> Solver f (KimchiConstraint f) a b
+  -> a
```
martyall, line 35: change this to `Gen a` and run in Effect

packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
```diff
@@ -527,6 +532,49 @@ type SrsBundle =
...
+-- | Regenerate the single-gate witness fixtures for the Lean checker, reusing the harness's
+-- | own gate circuits and concrete inputs. A no-op unless KIMCHI_WITNESS_DUMP_DIR is set (see
+-- | Test.Pickles.CircuitDiffs.WitnessDump), so ordinary diff runs are unaffected. The file names
+-- | match `formal/fixtures/*.json`, so a dump overwrites the committed fixtures in place.
+dumpGateFixtures :: Effect Unit
+dumpGateFixtures = witnessDumpDir >>= case _ of
+  Nothing -> pure unit
+  Just _ -> do
+    let
+      toPt :: Pallas.G -> AffinePoint Fp
```
martyall: This all disappears with `Gen a`

kejace:
This PR now includes

- generation of witness into existing `result/` JSON from pickles-circuit-diff test suite, using custom effect matcher 
- added the first gate so that we can see how it shows up in the test, with the corresponding lean code that validates it

packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
```diff
@@ -448,21 +447,60 @@ exactMatchEff
...
+-- | comparison JSON written to `circuits/results/`. Extracting a committed
+-- | `formal/fixtures/circuits/*.json` is `jq .witness circuits/results/<name>.json`.
+exactMatchWitnessEff
+  :: forall @a @b avar bvar
+   . CircuitType Fp a avar
+  => CircuitType Fp b bvar
+  => CheckedType Fp (KimchiConstraint Fp) avar
+  => String
+  -> (forall r. avar -> Snarky Fp (KimchiConstraint Fp) r bvar)
+  -> a
```
martyall: I think it's going to come out cleaner in the end to use `Gen a`

packages/pickles-circuit-diffs-types/src/Pickles/CircuitDiffs/Types.purs
```diff
@@ -20,9 +21,23 @@ type ComparableCircuit =
   , cachedConstants :: Array { variable :: Int, varType :: String, value :: String }
   }
 
+-- | A solved circuit-satisfaction instance for the PureScript circuit: the gate list and the
```
martyall: remove this comment

packages/pickles-circuit-diffs-types/src/Pickles/CircuitDiffs/Types.purs
```diff
@@ -20,9 +21,23 @@ type ComparableCircuit =
...
+-- | 15 register columns, all field elements as LE-hex strings. Self-contained in the shape the
+-- | Lean checker ingests (`formal/Kimchi/Json.lean`) — extracting `.witness` from a results
+-- | JSON yields a `formal/fixtures/circuits/*.json` verbatim.
+type WitnessExport =
```
martyall: What would it look like if 

```purescript

type WitnessExport = { witness :: Array (Array String), publicInputs :: Array String} 

type ComparableCircuit =
  { publicInputSize :: Int
  , gates :: Array ComparableGate
  , cachedConstants :: Array { variable :: Int, varType :: String, value :: String }
  , witness :: Maybe WitnessExport
  }
```

packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
```diff
@@ -448,21 +447,60 @@ exactMatchEff
...
+  -> SpecT Aff Unit Aff Unit
+exactMatchWitnessEff name circuit input =
+  exactMatchWith name do
+    builtState <- compile @Fp noAdvice (Proxy @a) (Proxy @b) (Proxy @(KimchiConstraint Fp))
+      circuit
+    let
+      solver =
+        makeSolver (Proxy @(KimchiConstraint Fp)) circuit
+          :: Solver Fp (KimchiConstraint Fp) a b
+    export <- buildWitnessExport builtState solver input
```
martyall: There is duplication within `buildWitnessExport` `fromCompiledCircuit` , eg they both call `makeGateData`. If `buildWitnessExport` only does witness generation and you make the type change suggested for `WitnessExport`, I think this will be more natural.

packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/WitnessDump.purs
```diff
@@ -0,0 +1,96 @@
...
+  runSolver solver input >>= case _ of
+    Left e -> unsafeCrashWith ("witness export: solver failed: " <> show e)
+    Right (Tuple _out assignments) -> do
+      let
+        { witness, publicInputs } = makeWitness
+          { assignments
+          , constraints: map _.variables gd.constraints
+          , publicInputs: builtState.publicInputs
+          }
+        gates = Array.mapWithIndex
```
martyall: suggestion: let `fromCompiledCircuit` handle this part

formal/Main.lean
```diff
@@ -1,7 +1,42 @@
...
-open Kimchi.Gate
+/-! Ingest a dumped `(circuit, witness)` and run the verified plonkish checker over the Pasta
+    base field. The object both the PureScript prover and the Lean proofs refer to is the
+    dumped gate list, and `Kimchi.Circuit.check` (faithful by `check_iff`) decides its
+    satisfaction.
 
-def main : IO Unit := do
-  IO.println s!"satisfies [egGood] = {satisfies [egGood]}"  -- true
-  IO.println s!"satisfies [egBad]  = {satisfies [egBad]}"   -- false
+    The witness source is the circuit-diff harness: `exactMatchWitnessEff` registrations embed
```
martyall: remove this whole comment. usage sould be documented in some readme or more likely some `tool` readme

formal/Kimchi/Field/Pasta.lean
```diff
@@ -0,0 +1,58 @@
+import Mathlib.Data.ZMod.Basic
```
martyall: This module is not needed, you should use ` CompElliptic.Fields.Pasta` (we are using this all over, claude will know what to do)

formal/Kimchi/Json.lean
```diff
@@ -0,0 +1,126 @@
+import Lean.Data.Json
```
martyall: we should namespace this under PS because we are also parsing fictures from rust. While they might coincidentally share parsers, it's not worth mixing them all in.

martyall (commented):
Many of the types that you want to parse into have already been defined in thr process of building out rust fixture parsers. If possible these kinds of types should be shared, and the only thing that should be different is the parsers themselves. they should ne namespaced PS vs rust 

also it should still be gated with env var cause for normal usage of yhis test suite we dont want it

martyall:
Rebuilt this branch on current main (was 7 commits behind, predating the A1–A3 index work) and applied the review — force-pushed as two commits:

**PS (`8ed6b3ae`)**
- `WitnessExport = { witness, publicInputs }`, a `Maybe` field on `ComparableCircuit` — the gate list is no longer serialized three times per results JSON
- witness generation is env-gated (`CIRCUIT_DIFFS_WITNESS_EXPORT=1`) and samples its input from a `Gen a` (`CIRCUIT_DIFFS_WITNESS_SEED`, default 42, logged) — without the flag the harness is byte-identical to before
- `fromCompiledCircuit` split into `gateDataOf` + `fromGateData`; `buildWitnessExport` only does witness generation (one `makeGateData` per registration)

**Lean (`2df4b21a`)**
- dropped `Kimchi/Circuit.lean`, `Checker/*`, `Field/Pasta.lean`, and `Main.lean` changes — main's A2 (`Kimchi.Index.Satisfies`) already provides the model and the derived checker, and `CompElliptic.Fields.Pasta` (vendored, certificate-verified) replaces the primality axiom
- new `Kimchi/Fixture/PS.lean`: PS-namespaced decoders (LE-hex + signed decimal) composing over the existing `Kimchi.Fixture` combinators; ingestion pads the dump to a two-power domain, synthesizes ω and the coset shifts, and constructs the index via `Index.build?` — every synthesized law is *decided*, so nothing is trusted
- driver follows the `scripts/check_*.{lean,sh}` convention (`check_ps_witness`); `make lean-check-witnesses` runs it

Verified end-to-end: `CIRCUIT_DIFFS_WITNESS_EXPORT=1 spago test … --example add_complete_step_circuit` → `bash formal/scripts/check_ps_witness.sh` gives `Satisfies` ✓ plus rejection of a corrupted gate cell, wired cell, and public input. Old tip was `5330b1e5` if you need to diff against it.

martyall:
Follow-up (`0fda73a4`, amends the Lean commit): the library's `checks`/`checksAt` pair was ad-hoc — a test report type in a library module, split only to dodge a do-notation parse issue. Now `Kimchi.Fixture.PS` exposes exactly one thing: `build : Raw → Except String Instance`, where `Instance` packs the padded domain, the decided index, the public input, and the witness table. The script states the propositions itself (`decide (Satisfies inst.idx …)` + the corruption rejections), in the manner of `check_index_fixture`. Re-verified end-to-end.

martyall:
Added the CI wiring as a third commit (`a83f6897`) rather than a stacked PR: `test-circuit-diffs` now exports the witness and hands it to a new `check-ps-witness` job via a run-scoped artifact; the Lean setup is factored into `.github/actions/setup-lean` (shared with lean.yml, plus a `formal/.lake/build` cache); and the two fixture scripts that had landed without CI coverage (`check_perm_fixture`, `check_index_fixture`) are wired into lean.yml. This PR's own CI run exercises the full chain.

PR #194, opened 2026-07-06
formal: executable IPA verifier over both Pasta curves — transcript re-derivation from wire data

martyall:
The executable IPA verifier, cross-validated against kimchi/proof-systems **over both Pasta curves**: given **wire data alone** — SRS, commitments, evaluation points, claimed evaluations, combination scalars, opening proof — `Kimchi.Verifier.Ipa.verify` re-derives the `U` base and every Fiat-Shamir challenge through the sponge layer (#195) and checks the two IPA acceptance equations. **Nothing transcript-derived is shipped**: the fixtures are exactly the wire format, and Lean derives the entire transcript on its own end.

## Single-index genericity

The verifier is generic over one `CommitmentCurve` bundle — the Lean analogue of the Rust `G: CommitmentCurve` associated types: base/scalar cardinalities with their field instances (`Fq`/`Fr` as projections), sponge spec, curve coefficient, map-to-curve. `Proof C` / `Input C` / `verify C` are indexed by the bundle, so a proof's type names its curve; `IpaVesta` and `IpaPallas` are each a bundle plus abbreviations. Everything modulus-dependent is **selected from the types**:

- the absorbed-scalar encoding `shift_scalar`: `Shifted_value` **Type1** (`(x−2²⁵⁵−1)/2`, Vesta) vs **Type2** (`x−2²⁵⁵`, Pallas) by the `n1 < n2` comparison — `Kimchi/Shifted.lean` now owns both directions with round-trip lemmas;
- the sponge's `absorb_fr` branch (direct embed vs high-bits/low-bit split, from #195).

The SvdW map generalizes identically (`Kimchi/Sponge/GroupMap.lean`): generic core + per-curve `Spec`s of proved numerals. The Pallas `setup()` root choice lands on the *other* primitive cube root — `pallas_endo²`, validated by `decide` — with production `to_group` vectors for both curves.

## What the check proves operationally

`scripts/check_ipa_fixture.sh` (CI-wired): four fixtures (single opening + 2×2 batch, per curve) parse as wire fields only, every point validated on its curve, and the verifier **ACCEPTS**; a corrupted claimed evaluation **REJECTS** (per curve). Acceptance is decided end-to-end: Poseidon permutation → duplex automaton → limb buffer and endo expansion → SvdW map → the `SRS::verify` schedule → the acceptance equations. The Pallas run exercises every branch Vesta cannot (Type2 shift, split `absorb_fr`, `fp_kimchi` parameters, `pallas_lam` expansion, opposite-root group map).

## Fixture provenance (thin-wrapper contract)

`tools/fixture-dump/ipa_dump`, curve-generic: proofs from the production `SRS::commit`/`SRS::open`, the production batched `SRS::verify` asserted at dump time, harness from poly-commitment's own `tests/ipa_commitment.rs::test_opening_proof`, deterministic (seeded ChaCha20). No reimplementation of anything.

Axiom surface unchanged: all 59 roots reduce to the allowed set.

Next (tracked in #196): the FS-validity axiom stated at the junction with `batch_soundness`, and the executable-verifier ↔ `BatchAccepts` reflection — together: *accepted wire data ⇒ genuine openings, under one named axiom plus the binding hypothesis*.


martyall:
Re-scoped: this check will be rebuilt on top of the sponge Fiat-Shamir instantiation (Poseidon permutation + FqSponge semantics + endo-expansion + group_map, modeled definitionally in Lean; 'Poseidon is a valid FS transform' becomes an explicitly flagged project axiom). Final fixture format carries only the proof and public inputs — Lean derives all challenges. The dumper and the extracted challenge values in the current fixtures survive as test vectors for validating the sponge stages during development. Draft until the sponge stack lands underneath.

formal/Kimchi/Verifier/Ipa.lean
```diff
@@ -0,0 +1,245 @@
...
+
+open CompElliptic.CurveForms.ShortWeierstrass
+open Kimchi.Sponge Kimchi.Sponge.FqSponge Kimchi.Commitment.IPA
+
+/-- The per-curve data of the verifier, bundled as a single index — base and scalar
+cardinalities with their field instances, the Fq-sponge spec, the curve coefficient `A`
+(for the complete addition), and the map-to-curve (projected to raw pairs). The field
+types are recovered as the projections `Fq`/`Fr`. -/
+structure CommitmentCurve where
+  q : ℕ
```
martyall: why p and q? Is this standard? Fr Fq is more standard  imo, or base and scalar field.

formal/Kimchi/Verifier/Ipa.lean
```diff
@@ -0,0 +1,245 @@
...
+  p : ℕ
+  [fieldQ : Field (ZMod q)]
+  [fieldP : Field (ZMod p)]
+  sponge : FqSponge.Spec q p
+  curveA : ZMod q
+  toGroup : ZMod q → ZMod q × ZMod q
```
martyall: Why not use SWCurve as the return type, that's what our group map actually uses no?

formal/Kimchi/Verifier/Ipa.lean
```diff
@@ -0,0 +1,245 @@
...
+attribute [instance] CommitmentCurve.fieldQ CommitmentCurve.fieldP
+
+/-- The base field. -/
+abbrev CommitmentCurve.Fq (C : CommitmentCurve) := ZMod C.q
+
+/-- The scalar field. -/
+abbrev CommitmentCurve.Fr (C : CommitmentCurve) := ZMod C.p
+
+/-- A point in CompElliptic's raw affine encoding: `(0, 0)` is the identity. -/
+abbrev Pt (C : CommitmentCurve) := C.Fq × C.Fq
```
martyall: Don't reinvent this type and redefine all of it's methods, use SWCurve (or AffinePoint) and you will inhereit all the mthods for free

formal/Kimchi/Verifier/Ipa.lean
```diff
@@ -0,0 +1,245 @@
...
+  z1 : C.Fr
+  z2 : C.Fr
+  sg : Pt C
+
+/-- A batched opening claim, uncombined — the verifier's wire input: the SRS, the
+per-polynomial commitments (one chunk each), the evaluation points, the claimed evaluation
+matrix (`evals[i][j]` = polynomial `i` at point `j`), the combination scalars, and the
+proof. -/
+structure Input (C : CommitmentCurve) where
+  srsG : Array (Pt C)
```
martyall: The SRS (including Gs and H) is not part of this. Curry whatever takes this, the SRS should always be supplised as a separate arg (see literally every other module which uses an SRS i this entire repo)

formal/Kimchi/Verifier/Ipa.lean
```diff
@@ -0,0 +1,245 @@
...
+  (cs.foldl (fun (acc : Pt C × C.Fr) P => (padd C acc.1 (psmul C acc.2 P), acc.2 * ξ))
+    (((0, 0) : Pt C), (1 : C.Fr))).1
+
+/-- The transcript encoding of an absorbed scalar (`shift_scalar`,
+`poly-commitment/src/commitment.rs`): the `Shifted_value` register form at the
+scalar-modulus bit size (255 for both Pasta fields) — Type1 (`(x − 2²⁵⁵ − 1)/2`) when the
+scalar modulus is below the base modulus, the Type2 shift (`x − 2²⁵⁵`) otherwise. The
+branch is the Rust `n1 < n2`, decided here from the cardinalities. -/
+def shiftScalar (q : ℕ) {p : ℕ} [Field (ZMod p)] (x : ZMod p) : ZMod p :=
+  if p < q then Kimchi.Shifted.shiftType1 255 x else Kimchi.Shifted.shiftType2 255 x
```
martyall: 255 is a magic number ?

formal/Kimchi/Verifier/Ipa.lean
martyall: Renamed: the bundle's cardinality fields are `base`/`scalar` (they are naturals — cardinalities — not the fields themselves); the field types are recovered as the projections `C.Fq`/`C.Fr`. (`q p` came from the merged sponge layer's generic parameters; that layer gets the same rename in a follow-up PR on main.)

formal/Kimchi/Verifier/Ipa.lean
martyall: Done — the bundle now carries `E : SWCurve (ZMod base)` and `toGroup : ZMod base → SWPoint E`, so `GroupMapVesta.toGroup` slots in directly with no projection anywhere, and the separate `curveA` field is gone (it's `E.A`).

formal/Kimchi/Verifier/Ipa.lean
martyall: Done, and you were more right than I knew: `SWPoint` already has `Add`/`Zero`/`Neg`/`DecidableEq`, and its `AddCommGroup` instance defines `nsmul := binNsmul` — the efficient binary scalar action — so even the double-and-add I hand-rolled was redundant. `Pt`/`padd`/`psmulNat`/`psmul`/`pmsm` are deleted; the verifier body is plain group notation (`c.val • Q + δ = …`) with `decide` on the inherited equality, and the only remaining local def is `msm := ∑ i, (a i).val • g i`, the visible mirror of `commitGen`. (The `.val •` ℕ-actions become `C.Fr`-actions once the reflection arc provides `Module C.Fr Point` — at which point this verifier is nearly literally the `Prop`-level verifier.)

formal/Kimchi/Verifier/Ipa.lean
martyall: Done — `Input` carries only the claim; `verify` takes the library's `Kimchi.Commitment.IPA.SRS C.Point` as a separate argument (`σ.k` replaces the round-count guard). One deliberate honesty note in the docstring: the library SRS bundles the randomisation base `U`, but the deployed protocol derives `U` from the transcript — `verify` derives it and never reads `σ.U`. That mismatch is precisely where the Fiat-Shamir assumption will connect the derived base to the abstract one.

formal/Kimchi/Verifier/Ipa.lean
martyall: Gone — the width derives from the bundle: `Nat.size C.scalar`, which is exactly arkworks' `MODULUS_BIT_SIZE` (255 for both Pasta scalar fields, but nowhere written). Both the Type1/Type2 branch (`scalar < base`, the Rust `n1 < n2`) and the bit size now flow from the cardinalities.

PR #205, opened 2026-07-07
formal: phase A1 — the kimchi index as data

martyall:
**Draft for representation review before anything consumes it** — per the phase-A ground rules (one stored representation; views bridged at definition; no hand-written executable twins).

`Kimchi/Index/Basic.lean`:

- **Stored data** (the only copy): `GateType` (the six formalized + `zero`), `GateRow` (type, `coeffs : Fin 15 → F`, `wires : Fin 7 → Fin 7 × Fin n` — kimchi's cyclic-successor wiring encoding), `Index` (gate table, `publicCount`, `zkRows`, `omega`, `endo`, `shifts`).
- **`Index.WF`** (`Prop` bundle over the stored data): primitive generator, zk/public row bounds, `CosetShifts`, wiring bijective + region-preserving. The generator and shift conditions are decidable on concrete data via the existing `Wiring.lean` certificates.
- **Derived views, each bridged where defined**: computable row forms (`selectorRow` — with booleanness a lemma, discharging the quotient layer's `hsel` hypothesis definitionally; `coeffRow`; `sigmaAddrRow`) → `columnPoly` interpolants (`eval_*Poly` bridges) → and `wiringPerm := Equiv.ofBijective` with definitional application (`wiringPerm_apply : rfl`), whose sigma interpolants coincide with `Permutation.sigmaPoly` by `rfl` (`sigmaPoly_eq_wiring`).

Everything compiles with zero sorries; the `rfl`-tightness of the two structural bridges is the evidence the representation carries no seams. Not yet here (by design, pending this review): the satisfiability predicate + `Decidable` instance (A2), the `index_dump` fixture (A2's dumper covers both), the quotient-bridge restatements (A3).

Also folds in the parked citation pass on the FS axioms.

**Review focus**: the stored types. `coeffs` as total `Fin 15 → F` (kimchi pads); wiring as the successor map with bijectivity in `WF` rather than a stored `Equiv.Perm`; `omega`/`endo`/`shifts` as index fields with their facts in `WF` (values with per-value facts — not instance fields, unlike the `CommitmentCurve` cardinalities).

### DIFF
--- a/formal/Kimchi/Index/Basic.lean
+++ b/formal/Kimchi/Index/Basic.lean
@@ -1,0 +1,178 @@
+import Kimchi.Quotient.Wiring
+
+/-!
+# The kimchi index: the circuit as data
+
+A kimchi circuit as the constraint system carries it (`ConstraintSystem`,
+proof-systems `circuits/constraints.rs`): a gate table — one `GateRow` per domain row,
+with its gate type, its fifteen coefficient cells, and its seven wire pointers — plus the
+domain generator, the coset shifts, the base-field endomorphism coefficient, and the
+zero-knowledge and public-input row counts. The index carries its laws: a value of
+`Index F n` is wellformed by construction (primitive generator, coset shifts, bounded
+row regions, a region-preserving bijective wiring), in the manner of `SWPoint`. On
+concrete data every law is decidable — the generator and shift conditions through the
+certificates of `Kimchi/Quotient/Wiring.lean`, the rest by `Fintype` instances — so
+parsers construct indices by deciding, never by trusting.
+
+**One stored representation.** The table is `Fin`-indexed data; the satisfiability
+predicate (milestone A2) and every proof consume it directly. Everything else is a
+*derived view* with its bridge proved at the definition site:
+
+* the **coefficient table** (`coeffTable`) — the `qTab` that the quotient layer's
+  `rowEnv` consumes: a `GateRow`'s coefficients become `ArgumentEnv.coeff` there. The
+  stored row and the evaluation environment share nothing else: `ArgumentEnv` is a
+  per-row view of witness *and* coefficients at evaluation time; `GateRow` is the static
+  circuit datum (type, coefficients, wiring);
+* the **row forms** (`selectorRow`, `coeffRow`, `sigmaAddrRow`) — computable functions
+  read off the table: the 0/1 indicator of a gate type, the coefficient columns, and the
+  wired-to addresses of the permutation;
+* the **interpolants** (`selectorPoly`, `coeffPoly`, `sigmaPoly`) — the `columnPoly`
+  images of the row forms, the polynomials the quotient layer consumes; each evaluates
+  back to its row form on the domain (`eval_selectorPoly` …), and the sigma interpolants
+  are exactly the wiring instantiation's (`sigmaPoly_eq_wiring`, definitional);
+* the **wiring permutation** (`wiringPerm`) — the stored successor map is kimchi's
+  encoding of the wiring (each cell points to the next cell of its copy cycle); it is a
+  permutation by the index's own law, and its underlying function is the stored map
+  (`wiringPerm_apply`, definitional).
+
+Gate types are the formalized six plus `zero` (no constraint — padding and wiring-only
+rows). Flagged optional gates (range check, foreign field, lookups) are out of scope.
+-/
+
+namespace Kimchi.Index
+
+open Polynomial Kimchi.Quotient Kimchi.Quotient.Permutation
+
+/-- The modeled gate types: the six formalized gates and the constraint-free `zero`. -/
+inductive GateType where
+  | zero
+  | generic
+  | poseidon
+  | completeAdd
+  | varBaseMul
+  | endoMul
+  | endoScalar
+  deriving DecidableEq, Repr
+
+/-- One row of the gate table: the gate type, the fifteen coefficient cells, and the
+seven wire pointers (each permuted cell names the next cell of its copy cycle —
+kimchi's cyclic-successor encoding of the wiring). The coefficients feed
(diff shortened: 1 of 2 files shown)


formal/Kimchi/Index/Basic.lean
```diff
@@ -0,0 +1,164 @@
...
+  | completeAdd
+  | varBaseMul
+  | endoMul
+  | endoScalar
+  deriving DecidableEq, Repr
+
+/-- One row of the gate table: the gate type, the fifteen coefficient cells, and the
+seven wire pointers (each permuted cell names the next cell of its copy cycle —
+kimchi's cyclic-successor encoding of the wiring). -/
+structure GateRow (F : Type*) (n : ℕ) where
```
martyall, line 61: isn't this ArgumentEnv? Don't we already have this?

formal/Kimchi/Index/Basic.lean
```diff
@@ -0,0 +1,164 @@
...
+  wires : Fin 7 → Fin 7 × Fin n
+
+/-- The index: the gate table and the domain/permutation constants
+(`ConstraintSystem`). -/
+structure Index (F : Type*) (n : ℕ) where
+  gates : Fin n → GateRow F n
+  publicCount : ℕ
+  zkRows : ℕ
+  omega : F
+  endo : F
```
martyall: what kind of endo? lambda or beta?

formal/Kimchi/Index/Basic.lean
```diff
@@ -0,0 +1,164 @@
...
+variable {F : Type*} [Field F] {n : ℕ}
+
+/-- The wiring as a map on cells: the stored successor pointers, read off the table. -/
+def wiringMap (idx : Index F n) (c : Fin 7 × Fin n) : Fin 7 × Fin n :=
+  (idx.gates c.2).wires c.1
+
+/-- Wellformedness of an index: a primitive domain generator, positive zero-knowledge
+region fitting in the domain, public rows within the unmasked region, coset shifts, and
+a wiring that is a bijection never crossing the zero-knowledge boundary. -/
+structure WF (idx : Index F n) : Prop where
```
martyall: WF is a bad type name, dont abbreviate

formal/Kimchi/Verifier/Reflection.lean
```diff
@@ -209,7 +209,21 @@ combined eval vector: `FiatShamirTreeB` with the deployed acceptance
 `Ipa.verify … = true` as the antecedent. This is the project's declared assumption that
 the Poseidon sponge provides a valid Fiat-Shamir transform — it packages the
 rewinding/forking extraction and the random-oracle behaviour of the sponge. It is the
-sole non-standard axiom of the headline `ipaVesta_sound`. -/
+sole non-standard axiom of the headline `ipaVesta_sound`.
+
+The corresponding probabilistic statement is established: transcript-tree extraction for
```
martyall: Just remove this, this citation is not important right now

formal/Kimchi/Index/Basic.lean
martyall: Also it seems strange to me that we have all of this constraints / type level information separate from the Index itself? Would it make it unusable to have them wiht? Because it seems otherwise we have to carry around both

formal/Kimchi/Index/Basic.lean
martyall: Not the same object, though the coefficients are the genuine point of contact. `ArgumentEnv` is the per-row *evaluation environment* — `{witnessCurr, witnessNext, coeff}`, containing witness data, existing only when a constraint is evaluated at a row (built by `rowEnv`). `GateRow` is the *static circuit datum* — no witness, plus two things `ArgumentEnv` rightly lacks (the type tag and the wiring). Merging would put witness data in the index or wiring/type tags in the constraint environment. The overlap is now an explicit bridge: `Index.coeffTable` is the `qTab` that `rowEnv` consumes, so a row's stored coefficients become `ArgumentEnv.coeff` through the one existing definition — A2's dispatch will literally be `rowEnv wTab idx.coeffTable i`.

formal/Kimchi/Index/Basic.lean
martyall: β — the base-field cube root (`pallas_endo`/`vesta_endo` at Pasta), because that's what `EndoMul.Holds` consumes and what kimchi's `cs.endo` is. Renamed the field `endoBase` with a docstring saying exactly this, including that λ is challenge-expansion data living in the sponge `Spec.lam`, not index data. (This distinction is a standing project rule — it has bitten before.)

formal/Kimchi/Index/Basic.lean
martyall: Renamed away entirely — see the reply on the second thread here: the laws are now fields of `Index` itself, so there is no auxiliary type to name.

martyall: Restructured as you suggested: the laws are now fields of `Index` itself (the `SWPoint` pattern) — a value of `Index F n` is wellformed by construction, and every downstream signature drops the second argument. Fixture construction still works because every law is decidable on concrete data (the `Wiring.lean` certificates for generator/shifts, `Fintype` decidability for the wiring laws): the parser constructs by deciding, `parseSWPoint`-style. For the generated-verifier goal this is the right architecture: proof fields erase at extraction (the compiled record is exactly the data struct), and the decidable certificates — with their proved `of_certificate` implications — are precisely the input-validation layer a provably-sound generated verifier runs at deserialization.

formal/Kimchi/Verifier/Reflection.lean
martyall: Removed — the citation paragraph is out of this PR entirely; the reference list stays parked in the ledger for whenever it's wanted.

PR #212, opened 2026-07-09
ci: wire the Lean witness check + the unwired fixture scripts

martyall:
Stacked on #189 (retarget to main after it lands — the consumer job runs its `check_ps_witness.sh`).

## What this wires

1. **`check-ps-witness`** (new job in `test.yml`): `test-circuit-diffs` now runs with `CIRCUIT_DIFFS_WITNESS_EXPORT=1` and uploads the witness-carrying results JSONs as a run-scoped artifact; the new job (`needs: test-circuit-diffs`) restores the Lean build and runs `formal/scripts/check_ps_witness.sh` against them. Artifacts are scoped to the workflow run, so the checker can only ever see the witness its own commit generated — the PR's "no committed fixture, always fresh" property carries into CI.
2. **`check_perm_fixture` + `check_index_fixture`**: landed with the permutation milestone and A2 but were never added to `lean.yml` (its last step was `check_fq_sponge`). Now wired.

## Structure

- **`.github/actions/setup-lean`** — the Lean setup extracted verbatim from `lean.yml`, shared by both workflows, plus an `actions/cache` on `formal/.lake/build` (keyed on toolchain + lakefile + `formal/**/*.lean`, with a prefix `restore-key` for incremental rebuilds). On a cache hit the consumer job is a few minutes; it only rebuilds Kimchi when `formal/` changed — in parallel with `lean.yml` doing the same. Side effect: `lean.yml` itself stops re-checking every proof on unrelated reruns.
- The consumer lives in `test.yml` (not `lean.yml`) because `needs:`/artifacts only span one workflow file.
- The visualizer's copy of the results is stripped of witness columns before bundling (`jq del(.purescript.witness)`) — checker input, not Pages content.
- `deploy` is **not** gated on the new job; add it to `deploy.needs` if witness failures should block Pages.

## Known tuning risk

The `formal/.lake/build` cache across jobs is the one piece that may need a tweak: `lake update mathlib` regenerates the manifest each run, and lake's trace invalidation can occasionally rebuild more than expected. If it thrashes, the fix is committing/pinning the manifest restore order — a follow-up, not a design change.

### DIFF
--- a/.github/actions/setup-lean/action.yml
+++ b/.github/actions/setup-lean/action.yml
@@ -1,0 +1,54 @@
+name: Setup Lean (Kimchi formalization)
+description: >
+  elan (pinned toolchain) + Mathlib olean cloud cache + a build-directory cache for the
+  Kimchi library itself, then `lake build Kimchi`. Shared by the Lean workflow and the
+  witness-check job in the CI workflow so both build the library the same way.
+runs:
+  using: composite
+  steps:
+    - name: Install elan (reads the pinned toolchain from lean-toolchain)
+      shell: bash
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh \
+          | sh -s -- -y --default-toolchain none
+        echo "$HOME/.elan/bin" >> "$GITHUB_PATH"
+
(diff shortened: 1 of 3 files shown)


martyall:
Folding into #189 as a third commit instead.

PR #213, opened 2026-07-09
formal: phase B — satisfiability from the one quotient check

martyall:
## Summary

Phase B of the index formalization: the full kimchi aggregate family as index data, and the chain from the shape of kimchi's **one production quotient check** down to `Satisfies` — the A2 predicate the derived checker decides. Built on the sound endosclmul gate (post-#211, there is exactly one EndoMul constraint list to model — 12 constraints, matching the pinned `CONSTRAINTS`).

**The headline** (`Kimchi/Index/Aggregate.lean`):

```
satisfies_of_evalCheck : (injective (β,γ) grid of accumulators whose aggregated
  21+3-member families pass the derandomized eval-check against t·Z_H)
  → Satisfies idx pub wTab
```

## The chain, commit by commit

1. **The family** (`45c7e940`): `fullFamily` — kimchi's shared alpha pool (`linearization.rs` `constraints_expr`): 21 gate members, each the selector-weighted *cross-gate sum* at one alpha slot (`gateConstraints_length` pins every gate's transcription to its production `CONSTRAINTS` constant), the public interpolant subtracted in the generic α⁰ slot, plus the 3 permutation members.
2. **Separation** (`068dc48c`): selector one-hotness collapses the shared pool back to per-gate rows — `Z_H ∣` the 21 summed members forces every row's gate branch of `rowSatisfies`, through each gate's existing `Argument.bridge`. The generic slot-0 public subtraction lands on `withPublic` via the new `Gate.Generic.withPublic_holds_iff`.
3. **Alpha instantiation** (`f05bf668`): `dvd_of_evalCheck` at the 24-member family — the aggregated eval-check pins every member.
4. **Copy conjunct** (`be4437fa`): the permutation stack gains a divisibility-form core at every level (`soundness_of_dvd` → `copy_soundness_of_dvd` → `copy_soundness_wired_of_dvd` → `Index.copy_soundness_of_dvd`), with the existing eval-check theorems rerouted through it — no statement changes, no duplicated proofs. `copy_of_fullFamily_dvd` feeds the family's last three members into it.
5. **Construction laws** (`26506198`): three facts of kimchi's construction become `Index` laws decided by `build?` — `public_generic` (Pub rows first), `public_coeffs` (`coeffs[0]=1` over zeros, per `generic.rs`), `masked_identity` (zk rows identity-wired). The `hpubgen` hypothesis disappears. Both index-constructing fixture checks accept the extended `build?` on real data.
6. **Assembly** (`5be55e34`): whole-grid copy (unmasked + `masked_identity`), public pinning (`public_coeffs` collapses slot 0 to `wTab i 0 = pub i`), and the packaged headline.

## Verification

- Full `lake build`, style clean.
- **77 axiom roots** (the three Phase-B headlines registered) reduce to the unchanged allowed set — standard logic, the two Pasta Hasse bounds, the two CM eigenvalues, the declared FS assumptions. No `sorry`, no new axioms.
- `check_index_fixture` (production mixed-gate dump, n=32) and `check_ps_witness` (PS harness dump) both **decide the three new laws true** and stay green end-to-end.

## Boundaries

- The challenge grids are the library's standing deterministic surrogate for Fiat–Shamir randomness (injectivity hypotheses; the declared assumption boundary — nothing new here).
- Completeness (honest witness ⟹ divisible family) is future work.


formal/Kimchi/Index/Aggregate.lean
```diff
@@ -122,6 +122,177 @@ noncomputable def fullFamily (idx : Index F n) (pub : Fin idx.publicCount → F)
...
+/-- **Member divisibility pins every slot at every row**: the live gate's `k`-th
+constraint value is the public value in slot `0` and `0` beyond (slots past the gate's
+list are the zero polynomial, so nothing is asserted vacuously — they evaluate to `0`
+outright). -/
+theorem eval_gateConstraints_of_dvd (idx : Index F n) (pub : Fin idx.publicCount → F)
+    (wTab : Fin n → Fin 15 → F)
+    (hdvd : ∀ k, k < gateAlphaCount → zH F n ∣ idx.gateMember pub wTab k)
+    (i : Fin n) (k : ℕ) :
+    ((idx.gateConstraints wTab (idx.gates i).typ).getD k 0).eval (idx.omega ^ (i : ℕ))
+      = if k = 0 then pubAt idx pub i else 0 := by
```
martyall: This ends up looking weird -- it doesn't look like a vanishing equation unless you know that 

```
((idx.gateConstraints wTab (idx.gates i).typ).getD k 0).eval (idx.omega ^ (i : ℕ))
      - if k = 0 then pubAt idx pub i else 0
```

is the original gate member equation. I think it's because we always have this trailing 

```
 - if k = 0 then pubAt idx pub i else 0
```

in the formulas. This also just seems like a trivial application of evalGateMemeber, becuase you just proved that 

```
(idx.gateMember pub wTab k).eval (idx.omega ^ (i : ℕ))
      = ((idx.gateConstraints wTab (idx.gates i).typ).getD k 0).eval
          (idx.omega ^ (i : ℕ))
        - (if k = 0 then pubAt idx pub i else 0)

```

and so if the vanishing polynomial divides one it obviously divides the other, hence it's 0 on the domain

formal/Kimchi/Index/Aggregate.lean
```diff
@@ -122,6 +122,177 @@ noncomputable def fullFamily (idx : Index F n) (pub : Fin idx.publicCount → F)
...
+    rw [idx.eval_gateMember] at h0
+    exact sub_eq_zero.mp h0
+  · have hlen := idx.gateConstraints_length_le wTab (idx.gates i).typ
+    have hk' : gateAlphaCount ≤ k := Nat.le_of_not_lt hk
+    have hpos : 0 < gateAlphaCount := by norm_num [gateAlphaCount]
+    rw [List.getD_eq_default _ _ (by omega), eval_zero, if_neg (by omega)]
+
+/-- All row constraint values vanish, given the eval bridge for the gate's polynomial
+list and vanishing of every slot. -/
+theorem allZero_of_bridge {P : List (Polynomial F)} {L : List F} {x : F}
```
martyall: I see this is used several times, so I understand why you factored it out, but it's dumb

formal/Kimchi/Index/Aggregate.lean
```diff
@@ -322,6 +322,42 @@ theorem fullFamily_dvd_of_evalCheck (idx : Index F n) (pub : Fin idx.publicCount
...
     hCdeg htdeg hcheck
 
+open Kimchi.Quotient.Permutation in
+/-- The permutation members of the full family: entries `21 + s` are the three
+permutation constraints at the index's wiring data. -/
+theorem fullFamily_perm (idx : Index F n) (pub : Fin idx.publicCount → F)
+    (wTab : Fin n → Fin 15 → F) (z : Polynomial F) (β γ : F) (s : Fin 3) :
+    idx.fullFamily pub wTab z β γ
+        ⟨gateAlphaCount + (s : ℕ), Nat.add_lt_add_left s.isLt gateAlphaCount⟩
+      = Permutation.constraints idx.omega idx.zkRows z (idx.permWitnessPoly wTab)
```
martyall, line 314: The result of this theorem reads horribly when you inline the tactics like this

formal/Kimchi/Index/Aggregate.lean
martyall: Agree. Two problems compounding: the getD/ite RHS makes a vanishing statement read as bookkeeping, and the theorem is (for k < 21) a two-line corollary of eval_gateMember + zH_dvd_iff. Fix: split it — `eval_gateMember_eq_zero_of_dvd` (member evals vanish on the domain, the actual vanishing statement) and state the slot consequences as two lemmas, k = 0 (`… = pubAt`) and k ≠ 0 (`… = 0`), so neither carries the trailing ite. The out-of-range k ≥ 21 branch (getD default) folds into the k ≠ 0 lemma.

formal/Kimchi/Index/Aggregate.lean
martyall: Agree — it's glue between two interfaces I chose badly, not a lemma with content. If `eval_gateConstraints_of_dvd` concludes in ∀-mem form (`∀ E ∈ gateConstraints …, E.eval (ω^i) = 0`) instead of getD-indexed form, this dissolves into `simpa [← hb, List.forall_mem_map]` at each use site and the definition disappears.

formal/Kimchi/Index/Aggregate.lean
martyall: Agree — anchored proof terms don't belong in a statement. `Fin.natAdd gateAlphaCount s` is the standard embedding and is exactly this index, so the LHS becomes `idx.fullFamily pub wTab z β γ (Fin.natAdd gateAlphaCount s)`; the rowLast `⟨n − zkRows, by …⟩` mk appears in several statements already (fullFamily, copy_soundness_of_dvd) and deserves a named def (`idx.rowLast` or similar) with the proof inside, once, instead of an inline omega per site.

martyall:
Review applied in `9e709f6b`: the getD/ite slot statement and `allZero_of_bridge` are replaced by `gateConstraints_vanish_of_dvd` (∀-mem form) + `generic_holds_of_dvd`, with the getD bookkeeping confined to proofs; `Index.unmaskedEnd` names the recurring boundary-row mk (also cleans the Soundness.lean copy statements); `fullFamily_perm` indexes via `Fin.natAdd`. The five per-branch `hvan0` lambdas in the dispatch — the other getD eyesores — collapse to two lines each. 77 roots + both fixture checks unchanged.

PR #215, opened 2026-07-09
formal: phase C completeness — satisfiability characterized by the one quotient check

martyall:
The main completeness arc (C2 + C3), on top of #214's prep. **The headline**:

```
fullFamily_dvd_of_satisfies :
  Satisfies idx pub wTab → ∀ β γ, Nondegenerate … β γ →
    ∃ z, ∀ s, Z_H ∣ idx.fullFamily pub wTab z β γ s
```

With Phase B's `satisfies_of_fullFamily_dvd`, satisfiability at a wellformed index is now **characterized** by the shape of kimchi's one quotient check — the round-trip that also audits every transcription in the aggregate family (a wrong sign or slot anywhere would make one direction unprovable). Note the deliberate asymmetry: soundness needs a challenge *grid*; completeness is *pointwise* per (β, γ) — the honest prover works at whatever challenges it's given, outside the degenerate locus.

## The chain

1. **C2a** (`Accumulator.lean`): `accumulator_of_prod_eq` — the converse telescoping: nonzero denominators + agreeing grand products yield the running-ratio column, pinned to 1 at both ends. The one place the nonzero-denominator hypothesis is genuinely needed (soundness was division-free).
2. **C2b** (`Wiring.lean`): `Nondegenerate` — no σ-side factor vanishes on the unmasked region; the formal rendering of the prover's division-by-zero edge case, with the degenerate (β, γ) confined to ≤ `7·(n − zkRows)` affine lines (the small bad locus a Fiat–Shamir sample misses, same currency as the soundness grids). And `prod_shiftSide_eq_prod_sigmaSide` — the grand products agree on a copy-invariant witness by *pure reindexing* (`Equiv.prod_comp`): the σ-side factor at `c` is the shift-side factor at `σ c`. No grid, no Vandermonde.
3. **C2c** (`Permutation.lean`): `zkpm_eval_zero` (the completeness twin of `zkpm_eval_ne_zero`) and `constraints_dvd_of_prods` — the honest accumulator interpolated through the domain, held constant on the mask (where zkpm erases the recurrence anyway), makes all three permutation constraints divisible. Converse of `soundness_of_dvd`.
4. **C3** (`Index/Aggregate.lean`): `permConstraints_dvd_of_copy` (index-level C2), `fullFamily_gate` (naming the `dif_pos` projection), the headline, and `exists_quotient_of_satisfies` — completeness in the production shape: honest `z` and, per fold challenge, a quotient `t` with `aggregate = t·Z_H` as a **polynomial identity**, so the eval-check passes at every node, not merely sampled ones.

## Verification

Full `lake build`, style clean, **80 axiom roots** (both completeness headlines registered) reduce to the unchanged allowed set; the index fixture check stays green.

One Lean note baked into the diff: `gateAlphaCount`/`permAlphaCount` are now `@[reducible]` — `Fin.natAdd`'s implicit index otherwise elaborates as `permAlphaCount` or `3` depending on context, defeq but `rw`-opaque.

## Not in scope (tracked)

Mask-invariance (the honest prover's *randomized* zk region satisfies `Satisfies`) — needs the boundary law for two-row gates; and the single-theorem iff (deriving grids from pointwise needs a field-cardinality bound). Both documented for the protocol-completeness phase.

### DIFF
--- a/formal/Kimchi/Index/Aggregate.lean
+++ b/formal/Kimchi/Index/Aggregate.lean
@@ -69,10 +69,10 @@ noncomputable def gateConstraints (idx : Index F n) (wTab : Fin n → Fin 15 →
 /-- The shared gate alpha-pool size: kimchi registers the pool at the largest gate's
 count — VarBaseMul's 21 (`linearization.rs`: "Only the max number of constraints
 matters"). -/
-def gateAlphaCount : ℕ := 21
+@[reducible] def gateAlphaCount : ℕ := 21
 
 /-- The permutation's alpha count (`permutation.rs`: `CONSTRAINTS = 3`). -/
-def permAlphaCount : ℕ := 3
+@[reducible] def permAlphaCount : ℕ := 3
 
 /-- The lengths of the gate transcriptions match kimchi's `CONSTRAINTS` constants. -/
 theorem gateConstraints_length (idx : Index F n) (wTab : Fin n → Fin 15 → F) :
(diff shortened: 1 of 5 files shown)


martyall:
Added the characterization iff (`satisfies_iff_fullFamily_dvd`) after discussion: `Satisfies ↔ honest quotient data at every nondegenerate (β, γ)`, under `(K+1)² ≤ |F|` (vacuous at Pasta; Fintype-only formulation by decision). The backward direction manufactures a nondegenerate grid — each cell forbids one γ per β, so the selection is `K+1` β's plus `K+1` γ's dodging `(K+1)·K` bad values (`exists_nondegenerate_grid`). 81 roots, same allowed set.

martyall:
Added mask-invariance after discussion, reformulated randomness-free as a **footprint theorem**: `satisfies_congr_unmasked` — tables agreeing off the mask are satisfiability-equivalent, because the four mask laws (`masked_zero`, the new `masked_boundary`, `masked_identity`+`wiring_region`, `public_le`) place the constraint system's entire read footprint in the unmasked region. `masked_boundary` (no two-row gate at the boundary row) is decided by `build?` and accepted by both fixtures; it deliberately narrows wellformedness — a circuit ending its unmasked region with a two-row gate has no mask-independent satisfiability at all. With this, the deferred list from the PR description is fully cleared: both items landed. 82 roots, same allowed set.

PR #217, opened 2026-07-09
formal: phase C milestone 1 — the chunk layer

martyall:
The ratified step zero of the Phase-C commitment layer: chunking, before anything is built at nc = 1 that would be unfaithful to production (the quotient `t` is multi-chunk in **every** real proof).

## The layer (`Kimchi/Commitment/IPA/Chunk.lean`)

Strictly additive over the nc = 1 argument — no existing statement changes, mirroring the verifier's own combine-then-open order:

- `chunkCoeffs` / `chunkPoly` — the width-`2^k` coefficient windows, as committed vectors and as degree-`< 2^k` polynomials;
- `eval_eq_sum_chunkPoly` — eval recombination: the reassembly `p = ∑ᵢ X^(i·2^k)·chunkᵢ` read at a point; kimchi's `combined_inner_product` identity;
- `commit_combine` — commitment recombination by linearity; kimchi's `chunk_commitment`;
- **`chunked_ipa_soundness`** (headline, axiom root #83): under commitment binding, an accepting IPA run on the `y = x^(2^k)`-power combination of honestly committed chunks of `p` (degree `< c·2^k`) forces the claimed value to be `p.eval x`. Binding pins the extracted witness to the combined chunk vector; recombination evaluates it.

The chunk count *is* the degree bound — this is the mechanism that will discharge Phase B's abstract `htdeg` hypotheses in the verifier-equation milestone.

## The adjudication

`ipa_dump` gains a chunked fixture (both curves): production `SRS::commit` in 3 chunks, production `chunk_commitment(y)`, an nc = 1 opening of the combined polynomial (computed coefficient-wise, mirroring `chunkCoeffs` exactly), accepted by production `SRS::verify`. `check_chunked_fixture` then re-derives the combination in Lean with the layer's own formulas — the `y`-power MSM of raw chunk points against the production combined commitment, the `y`-power fold of raw chunk evaluations against the claimed value — runs the executable verifier on the combined claim, and rejects a corruption. **Green on both curves**; wired into `lean.yml`. The four existing IPA fixtures regenerate byte-identically.

## Verification

Full `lake build`, style clean, **83 axiom roots** on the unchanged allowed set.

Next milestones per the phase plan: batch-with-chunks, then the verifier equation (stated chunk-aware from its first line), then the executable full verifier and FS composition.


martyall:
Review applied (`16241ffa`): the linearity is now first-class — `commitₗ : ((Fin (2^k) → F) × F) →ₗ[F] G`, with `commit_combine` demoted to the corollary mirroring kimchi's `chunk_commitment` formula (the image of `∑ᵢ yⁱ • (asᵢ, rsᵢ)` under the map, components read off pointwise). Statement unchanged, so the headline and fixture check are untouched; 83 roots, chunked adjudication still green on both curves.

formal/scripts/check_chunked_fixture.lean
```diff
@@ -0,0 +1,94 @@
...
+  raw chunk points equals the production combined commitment;
+* **eval recombination** (`eval_eq_sum_chunkPoly`'s formula): the `y`-power fold of the
+  raw chunk evaluations equals the claimed combined value;
+* **the executable verifier accepts** the combined claim (and rejects a corrupted one).
+-/
+
+open Lean Kimchi.Fixture Kimchi.Commitment.IPA Kimchi.Verifier
+open CompElliptic.CurveForms.ShortWeierstrass
+
+def parsePt (C : Ipa.CommitmentCurve) : Json → Except String C.Point :=
```
martyall: how do we not already have this?

formal/scripts/check_chunked_fixture.lean
```diff
@@ -0,0 +1,94 @@
...
+  parseSWPoint (parseZMod (n := C.base)) C.E
+
+def parseSRS (C : Ipa.CommitmentCurve) (j : Json) : Except String (SRS C.Point) := do
```
martyall: we must already have htis

formal/scripts/check_chunked_fixture.lean
```diff
@@ -0,0 +1,94 @@
...
+           evals := ← parseArrOf (parseArrOf parseS) (← fld "evals")
+           polyscale := ← parseS (← fld "polyscale")
+           evalscale := ← parseS (← fld "evalscale")
+           proof := { lr := ← parseArrOf (parsePair (parsePt C)) (← fld "lr")
+                      delta := ← parsePt C (← fld "delta")
+                      z1 := ← parseS (← fld "z1")
+                      z2 := ← parseS (← fld "z2")
+                      sg := ← parsePt C (← fld "sg") } }
+
+def checkFixture (C : Ipa.CommitmentCurve) (curveName : String) (path : String) :
```
martyall: why are we writing a separate check chunked fixture? All kimchi commitments are chunked? Why can't we ust modify the old check script to work on chunks of size 1, and hence use the same script we already have

tools/fixture-dump/src/bin/ipa_dump.rs
```diff
@@ -174,6 +174,144 @@ fn dump_fixture<P: SWCurveConfig>(
     );
 }
 
+/// Chunked fixture: one polynomial of degree `< c*N` committed in `c` chunks, the
+/// production `chunk_commitment(y)` combination at `y = x^N`, and an `nc = 1` opening of
+/// the combined polynomial `sum_i y^i * chunk_i` at `x`. The Lean side recombines the raw
+/// chunk points and chunk evaluations with its own formulas (`commit_combine` /
+/// `eval_eq_sum_chunkPoly`), compares against the production combination, and runs the
+/// executable verifier on the combined claim.
+fn dump_chunked_fixture<P: SWCurveConfig>(
```
martyall: Is there no way to unify this with the previous dumper and just control the number of chunks in our fixtures and make sure we have the breadth of test cases we need, specifically eg 1,2,3

formal/scripts/check_chunked_fixture.lean
martyall: Agree — duplicated from check_ipa_fixture.lean because scripts can't import each other, but that's copying the wart instead of fixing it: the IPA parse plumbing (parsePt/parseSRS/parseInput) should move to a library module (Kimchi/Fixture/Ipa.lean) that both scripts import, per the house pattern of scripts building on Kimchi.Fixture.

formal/scripts/check_chunked_fixture.lean
martyall: Agree — same fix: shared plumbing into Kimchi/Fixture/Ipa.lean.

formal/scripts/check_chunked_fixture.lean
martyall: Agree, and it's the deeper point — it's the same 'chunking is the type, not the edge case' observation from the scoping discussion: PolyComm is always a chunk vector and nc = 1 is just the one-element case, so the fixtures should mirror that. One unified schema (commitments as chunk arrays, evals as chunk lists), one check script whose recombination fold is the identity at c = 1. The recombination-vs-production comparison only has content for c > 1 but is harmless at c = 1. One honest caveat: this unification is schema/plumbing — the polyscale folding of chunks inside batched verification (mechanism (b)) remains milestone 2 regardless.

tools/fixture-dump/src/bin/ipa_dump.rs
martyall: Agree — one dump_fixture parameterized by (n_polys, n_points, n_chunks), with combine-then-open degenerating exactly to the plain path at n_chunks = 1. Matrix: 1×1×1, 2×2×1 (batch), 1×1×2, 1×1×3. One real constraint to encode: combine-then-open uses y = x^(2^k) at THE evaluation point, so n_chunks > 1 requires n_points = 1 for now (multi-point chunked batches are mechanism (b), milestone 2) — the unified function should assert that rather than silently accept the combination.

martyall:
Review applied (`ce893a40`): one dumper parameterized by (n_polys, n_points, n_chunks) with breadth 1×1×1, 2×2×1, 1×1×2, 1×1×3 per curve; one schema (commitments as chunk arrays, evals as chunk lists, plus the production-combined view); shared parse plumbing promoted to `Kimchi/Fixture/Ipa.lean`; `check_chunked_fixture` deleted — `check_ipa_fixture` runs the whole matrix with recombination-vs-production, verify, and corruption checks per fixture. Chunked fixtures assert single-point (the ζⁿ combination is anchored to the point; multi-point chunked batches are the polyscale mechanism, milestone 2). All eight fixtures green, 83 roots unchanged.

PR #218, opened 2026-07-10
formal: phase C milestone 2 — the chunked batch

martyall:
Phase C milestone 2: kimchi's **second** chunk-fold mechanism. Milestone 1 (#217) modeled combine-then-open — one polynomial's chunks folded by `x^(2^k)` powers into one commitment. This models what the deployed batch verification actually does with multi-chunk `PolyComm`s: **nothing is pre-combined** — each chunk is one *segment* of the batch stream, consumed polynomial-outer, chunk-inner, at one consecutive polyscale power per segment (`combined_inner_product` / `combine_commitments`, `poly-commitment/src/commitment.rs`).

## The observation that shapes the proof

The production segment layout is exactly the flat `nc = 1` batch under the flattening of `(polynomial, chunk)` pairs. So the milestone adds **no new extraction**:

* **`Batch.lean`** — the chunked combiners `chunkedCombinedCommitment` / `chunkedCombinedInnerProduct` over per-polynomial chunk counts `nc : Fin n → ℕ` (ragged, as in production: `t` has 7 chunks, most polys 1), with `segmentOffset` as the stream position; and the **flattening lemmas**: the chunked combiners are the plain combiners of the flattened segment family under `finSigmaFinEquiv`. Chunking changes the indexing, not the combination.
* **`Chunk.lean`** — the assembly inverse of windowing: `assemblePoly` builds the long polynomial with prescribed chunk windows (`chunkCoeffs_assemblePoly`, `assemblePoly_natDegree_lt`).
* **`Soundness/ChunkedBatch.lean`** — the headline **`chunked_batch_soundness`** (axiom root #84): an injective `(ξ, r)` acceptance grid — the polyscale rows now range over the *segment count* `∑ i, nc i` — plus the Fiat–Shamir tree hypothesis and DL-binding bind every polynomial of a chunked batch to **one genuine polynomial of degree `< nc i · 2^k`**: each wire chunk commits to a window of it, and the claimed chunk evaluations recombine (by `x^(2^k)` powers) to its true evaluations. Proof: flatten by the lemmas, extract one witness per segment via `batch_soundnessA`, assemble. Trust boundary identical to `Soundness/Batch.lean`.

This is the form milestone 3 (the verifier equation) consumes: every commitment in the real verifier's batch goes through this mechanism, and the degree bounds Phase B needs (`hCdeg`/`htdeg`) will be discharged from chunk counts + this binding.

## Production adjudication

New fixture kind (mechanism (b)): `ipa_chunked_batch_{vesta,pallas}.json` — 2 polys × 2 points × 2 chunks, opened by the production `SRS::open` **on the raw chunked polynomials** and accepted by `SRS::verify` consuming the multi-chunk `PolyComm`s as-is; the fixture records the production flat combination targets (`combine_commitments` at `rand_base = 1`, `combined_inner_product`). The eight existing fixtures regenerate byte-identical.

`check_ipa_fixture` gains three chunked-batch checks per curve:
1. the executable chunked combiners reproduce the production segment-stream targets;
2. the `nc = 1` executable verifier **accepts the flattened segment view** — the flattening lemma adjudicated end-to-end against a production chunked opening;
3. a corrupted chunk evaluation is rejected.

10/10 fixtures green; full `lake build` clean; 84 axiom roots reduce to the allowed set; style OK.


martyall:
Follow-up commit 11d2eff1: **ragged chunk counts**, per review discussion.

The dumper's `dump_chunked_batch_fixture` now takes per-polynomial chunk counts (`n_chunks: &[usize]`), matching the deployed batch, which mixes widths. New fixture `ipa_chunked_ragged_{vesta,pallas}.json`: 2 polys × 2 points with chunk counts **[1, 3]** — the case where `segmentOffset` being a genuine prefix sum (rather than `i · nc`) matters — opened by production `SRS::open` on the raw polynomials and accepted by `SRS::verify` as before.

The Lean side needed no code changes beyond adding the fixtures to the check loop: the theorem was already stated over ragged `nc : Fin n → ℕ`, and the executable combiners / `toFlatInput` already read each polynomial's own chunk-array length. 12/12 fixtures green, including the ragged flat-verify (the flattening lemma adjudicated end-to-end at unequal group widths).

The uniform `ipa_chunked_batch_*` fixtures regenerate with only the `num_chunks` schema field becoming a list (`[2, 2]`); all cryptographic content is byte-identical.

PR #220, opened 2026-07-10
[04/04] pickles-circuit-diffs: witness generation for the remaining four gates

kejace:
Final chunk of the #167 decomposition (see [01/13] … [04/13]). Turns on witness generation for the remaining four kimchi gates, so all seven are exercised end-to-end against the Lean index model.

## What

Flips `var_base_mul` / `endo_mul` / `endo_scalar` / `poseidon` from `exactMatchEff` to `exactMatchWitnessEff`, so — under `CIRCUIT_DIFFS_WITNESS_EXPORT` — each carries a seeded, solved witness on the PureScript side of its comparison JSON, exactly as `add_complete` already does. Two `Gen` helpers:

- `genScalar` — a positive `Int` scalar, so it sits well under `2^128` (valid for the `SizedF 128` decodes in endo-scalar / endo-mul) and clear of var-base-mul's small non-degeneracy forbidden band; seeded, so witnesses are reproducible.
- `genV3` — three field elements (the Poseidon permutation is total, so any input solves).

The now-unused `compileKFF` / `compileV3` helpers are removed (`compilePF` stays — `scale_fast2` uses it). Net `+18 / −25`.

## No Lean changes

`Index.rowSatisfies` already dispatches all seven gate types through the quotient layer's `rowWitness` transcriptions, and the `check-ps-witness` CI job already scans every `results/*.json` — so these four gates flow through the existing checker automatically. This PR is purely turning on their witnesses.

## Verified end-to-end

Locally, the exact pipeline CI runs:

```
CIRCUIT_DIFFS_WITNESS_EXPORT=1 make test  → 6/6 gate diffs pass, 4 new witnesses generated
scripts/check_ps_witness.sh               → all 5 gates:
   ✓ Satisfies (decided from the index-model predicate)
   ✓ corrupted gate cell / wired cell / public input each rejected
```

The corruption rejections confirm each gate's constraints are genuinely checked, and that the quotient-layer row layouts match the harness column layouts (109 rows for var_base_mul, 41 for endo_mul, 18 for poseidon, 12 for endo_scalar) — no mismatches.

With this, the #167 ingestion + verified-checker work is complete for all seven gates.

### DIFF
--- a/packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
+++ b/packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
@@ -89,6 +89,7 @@ import Snarky.Backend.Kimchi.Types (CRS)
 import Snarky.Circuit.CVar (add_) as CVar
 import Snarky.Circuit.DSL (class BasicSystem, class CheckedType, class CircuitType, BoolVar, F(..), FVar, SizedF, addConstraint, all_, and_, any_, assertEqual_, assertNonZero_, assertNotEqual_, assertSquare_, assert_, const_, div_, equals_, exists, if_, inv_, mul_, or_, pow_, unpack_, xor_)
 import Snarky.Circuit.DSL.Monad (Snarky)
+import Snarky.Circuit.DSL.SizedF (toField) as SzF
 import Snarky.Circuit.Kimchi.AddComplete (Finiteness(..), addFast)
 import Snarky.Circuit.Kimchi.EndoMul (endo)
 import Snarky.Circuit.Kimchi.EndoScalar (toField)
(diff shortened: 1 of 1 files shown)


packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
```diff
@@ -369,6 +349,16 @@ affinePt g = case toAffine g of
 genPallasPoint :: Gen (AffinePoint Fp)
 genPallasPoint = affinePt <<< power generator <$> chooseInt 1 top
 
+-- | A random scalar for the gate gadgets: a positive `Int`, so it sits well under 2^128
+-- | (valid for the `SizedF 128` decodes in endo-scalar / endo-mul) and clear of the small
+-- | non-degeneracy forbidden band of var-base-mul; seeded, so witnesses are reproducible.
+genScalar :: Gen (F Fp)
+genScalar = F <<< fromInt <$> chooseInt 16 top
```
martyall: you shouldn't need this, you should be able to just use arbitrary

packages/pickles-circuit-diffs/test/Test/Pickles/CircuitDiffs/Main.purs
```diff
@@ -369,6 +349,16 @@ affinePt g = case toAffine g of
...
+
+-- | Three random field elements (the Poseidon permutation is total, so any input solves).
+genV3 :: Gen V3
```
martyall: there should be an arbitreary instance for Vector of field elems

martyall (approved):

PR #226, opened 2026-07-10
formal: snarky DSL port, 2 — interpreters and the interpreter laws

kejace:
Stacked on #225. The two pure interpreters over the reified `CircuitM` tree, and the theorems that justify the deep embedding.

## Interpreters

- **`Snarky/Builder.lean`** — `build : CircuitM F c α → Nat → Built c α`, the PS `runCircuitBuilder`/`builderOps` without `Effect`/`Ref`s: allocates variables sequentially, collects constraints in emission order, and (like PS) never touches a witness payload. `Built` is a named-field result (`result`/`nextVar`/`constraints`), so downstream statements read `(build m nv).constraints` rather than `.2.2`. The constraint list is the DSL-level view — the natural first comparison point against a PS-side dump.
- **`Snarky/Prover.lean`** — `prove : … → Nat → Assignments F → Except EvalError (Proved F α)`, the PS `runCircuitProver`/`proverOps`: runs witness computations against the accumulating `Assignments`, checks each constraint with a caller-supplied `holds : c → Assignments F → Bool` as it is added (PS `proverConstraint`), and errors on any re-assignment — a deliberately *stronger* invariant than PS, which is what makes runs monotone. `Proved` mirrors `Built` field-for-field (`result`/`nextVar`/`assignments`), making the agreement laws read symmetrically.
- `CVar.lean` grows the assignment-extension order `Assignments.Le`, the guarded `extendPairs` (consuming a zipped `(variable, value)` pair list, so pairing allocation with witness values is total by construction — per review, the only failure mode is a genuine re-assignment), and `CVar.eval_le` (evaluation is stable under extension).

## The laws (`Snarky/Laws.lean`)

1. **`build_eraseWitness`** — witness-independence: `build` factors through `eraseWitness`, the function stripping every `AsProver` payload from the tree, so circuits with equal erasures build identical constraint systems (`build_eq_of_eraseWitness`; for literal circuits differing only in witness code the erasure equality is `rfl`). In PS this is an intention; here it is a structural induction — and per review, stated with an erasure *function* rather than an auxiliary shape relation.
2. **`prove_build_agrees`** — on a successful prover run, builder and prover compute the same result and allocate variables in lockstep.
3. **`prove_sound`** — completeness: if the prover run succeeds, the *final* assignment satisfies *every* constraint the builder emits (each constraint was checked at emission time; `prove_assignments_le` + monotonicity of `holds` carry the check to the end). Once `c` is instantiated at Kimchi gate rows, this is the bridge toward `Kimchi.Index.Satisfies`.

## Verification

- `lake build` clean, `scripts/check-style.sh` clean
- All five theorems wired into `scripts/check_axioms.lean`; closure is `[propext]` (no `Classical.choice`)

### DIFF
--- a/formal/Snarky/Laws.lean
+++ b/formal/Snarky/Laws.lean
@@ -1,0 +1,201 @@
+import Snarky.Prover
+
+/-!
+# The interpreter laws
+
+The theorems the deep embedding exists to state — none of them is expressible against the
+final-tagless PureScript original:
+
+1. **Witness-independence** (`build_eraseWitness`): the constraint builder never
+   inspects a witness payload — `build` factors through `eraseWitness`, the function
+   stripping every `AsProver` computation from the tree. Two circuits differing only in
+   their witness payloads have equal erasures, hence identical constraint systems
+   (`build_eq_of_eraseWitness`). This is structural: continuations receive only fresh
+   `Variable`s, never field values.
+2. **Interpreter agreement** (`prove_build_agrees`): on a successful prover run, the
+   builder and the prover compute the same result and allocate variables identically —
+   the deep-embedding counterpart of PS's builder/prover running the *same* closure
+   against two `CircuitOps` records.
+3. **Completeness** (`prove_sound`): if the prover run succeeds, the final assignment
+   satisfies *every* constraint the builder emits. The prover checks each constraint
+   against the assignment current at emission time; the final assignment only extends it
+   (`prove_assignments_le`), so a monotone `holds` stays true. This is the DSL-level
+   bridge that, once `c` is instantiated at Kimchi gate rows, will feed
+   `Kimchi.Index.Satisfies`.
+-/
+
+namespace Snarky
+
+variable {F c : Type u}
+
+/-! ## Witness-independence -/
+
+/-- Strip every witness payload from the tree, leaving the circuit's shape: the
+`AsProver` computations at `existsOp`/`assignOp` nodes are replaced by the trivially
+failing one. Two circuits differ only in their witness code exactly when their erasures
+are equal — for literal circuit terms that equality is `rfl`. -/
+def eraseWitness : CircuitM F c α → CircuitM F c α
+  | .pure a => .pure a
+  | .freshOp k => .freshOp fun v => eraseWitness (k v)
+  | .addConstraintOp con k => .addConstraintOp con (eraseWitness k)
+  | .existsOp n _ k =>
+    .existsOp n (fun _ => .error (.custom "erased")) fun vs => eraseWitness (k vs)
+  | .assignOp vs _ k =>
+    .assignOp vs (fun _ => .error (.custom "erased")) (eraseWitness k)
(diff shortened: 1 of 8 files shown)


formal/Snarky/Prover.lean
```diff
@@ -0,0 +1,68 @@
...
+witness computation succeeds, no variable is assigned twice, and every constraint holds
+when added. -/
+def prove (holds : c → Assignments F → Bool) :
+    CircuitM F c α → Nat → Assignments F → Except EvalError (Proved F α)
+  | .pure a, nv, env => .ok ⟨a, nv, env⟩
+  | .freshOp k, nv, env => prove holds (k nv) (nv + 1) env
+  | .addConstraintOp con k, nv, env =>
+    if holds con env then prove holds k nv env
+    else .error .unsatisfiedConstraint
+  | .existsOp n wit k, nv, env =>
```
martyall, line 52: It seems like `wit` should know the number of return variable-assignments which should should match `n`, ie it seems like `extend_vec` should never error on a length mismatch in practice

formal/Snarky/Laws.lean
```diff
@@ -0,0 +1,197 @@
...
+  | assignOp {n : Nat} (vs : Vector Variable n) (wit wit' : AsProver F (Vector F n)) {k k'}
+      (h : SameShape k k') :
+      SameShape (.assignOp vs wit k) (.assignOp vs wit' k')
+  | labelOp (s : String) {k k'} (h : SameShape k k') :
+      SameShape (.labelOp s k) (.labelOp s k')
+
+/-- **Witness-independence of the builder**: the constraint system (and the result, and
+the variable numbering) depends only on the shape of the circuit, never on the witness
+computations stored at `existsOp`/`assignOp` nodes. -/
+theorem build_eq_of_sameShape {m m' : CircuitM F c α} (h : SameShape m m') :
```
kejace: Is is possible to prove this without the `inductive SameShape` above?

formal/Snarky/Prover.lean
kejace: Right — the mismatch arm was dead code. Fixed: `extendVec` is now `extendPairs : Assignments F → List (Variable × F) → …`, consuming a zipped pair list (`(allocRange nv n).toList.zip xs.toList`), so pairing allocation with witness values is total by construction. The mismatch arm and its `.custom` error are gone; the only remaining failure mode is a genuine re-assignment (`conflict`). Bonus: `le_extendPairs` inducts over one list instead of two.

formal/Snarky/Laws.lean
kejace: Yes — and it's nicer. `SameShape` is deleted, replaced by an erasure *function* `eraseWitness` (strips every `AsProver` payload from the tree). Witness-independence is now `build_eraseWitness : build (eraseWitness m) n = build m n` — *build factors through erasure* — with the two-circuit version a one-line corollary (`build_eq_of_eraseWitness`). For literal circuits differing only in witness code, the erasure equality is `rfl`, so no relation and no relation-induction is needed.

PR #227, opened 2026-07-10
formal: snarky DSL port, 3 — the typed layer and an executable R1CS example

kejace:
Stacked on #226 (which stacks on #225). Completes the first milestone of the port.

## The typed layer (`Snarky/Types.lean`)

- **`CircuitType F val var`** — PS `CircuitType f a var`, with the field-vector size type-level (`Vector F size`) instead of PS's runtime `sizeInFields` contract.
- **`CheckedType`** — PS `CheckedType`'s `check`, emitted by `witness` under *both* interpreters.
- **`BasicSystem`** — PS `Snarky.Constraint.Basic` (`r1cs`, `boolean`), keeping the constraint type abstract.
- **`witness`** — PS `exists` (renamed; Lean keyword): allocate, record the witness computation, assemble the bundle, emit `check`. Plus `mul`/`assertEq`/`readVar` and `FVar`/`BoolVar` instances.

Deliberately not ported: the generic-deriving machinery (`GCheckedType`/`RCheckedType` — Lean would use `deriving`/macros) and the numeric-tower instances (sugar over the ported combinators).

## The example (`Snarky/Example.lean`)

A concrete R1CS backend over `Fin 17`: witness two values, multiply, assert the product. Every check closes by `decide`, so the file is a regression test for the executable semantics:

- the builder's constraint dump is checked literally: `[⟨.var 0, .var 1, .var 2⟩, ⟨.var 2, .const 1, .const 15⟩]`
- the prover accepts the honest run and rejects a falsified constant
- `prove_sound` is instantiated at the backend (monotonicity discharged once via `CVar.eval_le`)

## A trap worth knowing

Core `Vector.map` runs through `Array.mapM`, whose loop is compiled by **well-founded recursion** — opaque to kernel reduction, so it silently wedges every `decide` downstream of `witness`. `Snarky.mapVec` is the List-backed, definitionally transparent replacement; the gotcha is documented in `formal/CLAUDE.md`, which now also describes the `Snarky` library and scopes the "no circuit monad" remark to the `Kimchi.*` namespace.

## Module layout (per review)

Laid out along the PS namespacing rather than one `Types.lean`:

- **`Snarky/Types.lean`** — the `CircuitType`/`CheckedType` classes and `FVar`/`BoolVar` instances only (PS `Snarky.Circuit.Types`).
- **`Snarky/Constraint/Basic.lean`** — the `BasicSystem` backend interface (PS `Snarky.Constraint.Basic`).
- **`Snarky/Constraint/R1CS.lean`** — the plain R1CS backend, extracted from the example.
- **`Snarky/DSL.lean`** — the user-facing combinators: `witness` (PS `exists`; circuit-side, runs under both interpreters), `readVar` (prover-side), `mul`, `assertEq` — the PS `Snarky.Circuit.DSL` surface.
- **`Snarky/Vec.lean`** — kernel-reduction-friendly `Vector` utilities, with the full rationale in the module docstring (core `Vector.map` runs through `Array.mapM`, compiled by well-founded recursion, opaque to the kernel — one call wedges every downstream `decide`).
- `Snarky/Example.lean` keeps just the `Fin 17` circuit and its `decide`-checked runs.

## Verification

- full `lake build` clean (Kimchi + Snarky + demo), `scripts/check-style.sh` clean, `scripts/check_axioms.sh` green (90 roots)

### DIFF
--- a/formal/Snarky/Example.lean
+++ b/formal/Snarky/Example.lean
@@ -1,0 +1,59 @@
+import Snarky.DSL
+import Snarky.Constraint.R1CS
+import Snarky.Laws
+
+/-!
+# End-to-end example: an R1CS multiplication circuit
+
+The `Snarky.Constraint.R1CS` backend over `Fin 17`, exercising the whole stack: the typed
+`witness` combinator, both interpreters, and an instantiation of the completeness
+theorem. Everything here reduces by `rfl`/`decide`, so this file doubles as a regression
+test for the executable semantics.
+
+The circuit mirrors the smallest interesting PS example: witness two field elements,
+multiply them (one witnessed product + one `r1cs` constraint), and assert the result
+equals a constant.
+-/
+
+namespace Snarky.Example
+
+open Snarky Snarky.Constraint
+
+/-! ## The circuit -/
+
+abbrev F17 := Fin 17
+
+/-- Witness `x = 3` and `y = 5`, multiply, assert the product is `15`. -/
+def mulCircuit : CircuitM F17 (R1CS F17) (FVar F17) := do
+  let x ← witness (val := F17) (pure 3)
+  let y ← witness (val := F17) (pure 5)
+  let z ← mul x y
+  assertEq z (.const 15)
+  pure z
+
+/-! ## Running it -/
+
+/-- The builder allocates three variables (`x`, `y`, and the product). -/
+example : (build mulCircuit 0).nextVar = 3 := by decide
+
+/-- The builder emits two constraints (`x * y = z` and `z * 1 = 15`), in emission order. -/
+example : constraints mulCircuit =
+    [ ⟨.var 0, .var 1, .var 2⟩, ⟨.var 2, .const 1, .const 15⟩ ] := by decide
+
+/-- The prover succeeds: every witness computation runs and every constraint holds. -/
+example : (prove R1CS.holds mulCircuit 0 Assignments.empty).isOk = true := by decide
+
+/-- Changing the asserted constant makes the prover reject at the constraint check. -/
+example :
+    (prove R1CS.holds (do let z ← mulCircuit; assertEq z (.const 14))
+      0 Assignments.empty).isOk = false := by
+  decide
+
+/-- The completeness theorem, instantiated: any successful run of `mulCircuit` yields
+assignments satisfying every built constraint. -/
+example {x : FVar F17} {nv : Nat} {env : Assignments F17}
+    (h : prove R1CS.holds mulCircuit 0 Assignments.empty = .ok ⟨x, nv, env⟩) :
+    ∀ con ∈ constraints mulCircuit, con.holds env = true :=
+  prove_sound (holds := R1CS.holds) (fun _con _ _ hle hh => R1CS.holds_mono hle hh) h
+
+end Snarky.Example
--- a/formal/Snarky/Types.lean
+++ b/formal/Snarky/Types.lean
@@ -1,0 +1,73 @@
+import Snarky.Monad
+import Snarky.Vec
+import Snarky.Constraint.Basic
+
+/-!
+# Circuit types
+
+Port of the typed layer of the DSL: `CircuitType` (PS `class CircuitType f a var` from
+`Snarky.Circuit.Types`) and `CheckedType` (PS `class CheckedType f c var`,
+packages/snarky/src/Snarky/Circuit/DSL/Monad.purs lines 488–531), together with the
+`FVar`/`BoolVar` instances. The user-facing combinators built on these classes
+(`witness`, `readVar`, `mul`, `assertEq`) live in `Snarky.DSL`; the backend constraint
+interface lives in `Snarky.Constraint.Basic`.
+
+Improvement over PS: the field-vector size is type-level (`Vector F size` instead of
+`Array f` plus a runtime `sizeInFields` contract), so the length obligations disappear.
+
+Out of scope for this port (deliberately): the generic-deriving machinery
+(`GCheckedType`/`RCheckedType`) — Lean would grow `deriving` handlers or macros instead.
+-/
+
+namespace Snarky
+
+/-! ## The type classes -/
+
+/-- A bidirectional encoding of a value type `val` as `size` field elements, together with
+its variable-bundle counterpart `var` (PS `CircuitType f a var`). -/
+class CircuitType (F : Type u) (val : Type u) (var : outParam (Type u)) where
+  size : Nat
+  valueToFields : val → Vector F size
+  fieldsToValue : Vector F size → val
+  varToFields : var → Vector (CVar F) size
+  fieldsToVar : Vector (CVar F) size → var
+
+/-- Variable bundles whose well-formedness is enforced by constraints: `check` is emitted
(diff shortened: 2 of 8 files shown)


formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+
+/-! ## The `witness` combinator and friends -/
+
+variable {F c val var : Type u}
+
+/-- A List-backed `Vector.map`. Core's `Vector.map` runs through `Array.mapM`, whose loop
+is compiled by well-founded recursion and therefore opaque to kernel reduction — it would
+wedge every `decide`-based check downstream of `witness`. `List.map` is structural, so
+this version reduces. -/
+def mapVec (f : α → β) (v : Vector α n) : Vector β n :=
```
martyall: If vector is reimplemented, then these kind of utils need to be defined in that module

formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+  ⟨⟨v.toList.map f⟩, by simp⟩
+
```
kejace: Why do we need this?

formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+/-- Witness a typed value (PS `exists`, renamed — `exists` is a Lean keyword): allocate
+`size` fresh variables, record the witness computation for prover runs, assemble the
+variable bundle, and emit its `check` constraints under both interpreters. -/
+def witness [inst : CircuitType F val var] [CheckedType F c var]
```
martyall: seems like this belongs in some other module, ie prover specific

formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+    (compute : AsProver F val) : CircuitM F c var :=
+  .existsOp inst.size (fun env => (compute env).map inst.valueToFields) fun vs => do
+    let v := inst.fieldsToVar (mapVec CVar.var vs)
+    CheckedType.check (c := c) v
+    pure v
+
+/-- Read a typed variable bundle back to its value during a prover run (PS `read`). The
+length check is dynamic (it always succeeds) to keep the definition kernel-reducible
+without a `mapM`-length lemma. -/
+def readVar [Add F] [Mul F] [inst : CircuitType F val var] (v : var) : AsProver F val :=
```
martyall: seems like this belongs in some other module, ie prover specific

formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+  fun env => do
+    let fields ← (inst.varToFields v).toList.mapM (CVar.eval · env)
+    if h : fields.length = inst.size then
+      pure (inst.fieldsToValue ⟨⟨fields⟩, by simpa using h⟩)
+    else
+      .error (.custom "readVar: size mismatch")
+
+/-- Multiply two field variables: witness the product, constrain it with `r1cs`
+(PS `mul_`). -/
+def mul [Add F] [Mul F] [BasicSystem F c] (x y : FVar F) : CircuitM F c (FVar F) := do
```
martyall: these don't belong here, this is a types module

formal/Snarky/Types.lean
```diff
@@ -0,0 +1,126 @@
...
+  let z ← witness (val := F) do
+    let xv ← AsProver.readCVar x
+    let yv ← AsProver.readCVar y
+    pure (xv * yv)
+  addConstraint (BasicSystem.r1cs x y z)
+  pure z
+
+/-- Assert that two field variables are equal, via `x * 1 = y` (PS `assertEqual_`). -/
+def assertEq [One F] [BasicSystem F c] (x y : FVar F) : CircuitM F c PUnit :=
```
martyall: these don't beloing here, it's a types module

formal/Snarky/Example.lean
```diff
@@ -0,0 +1,90 @@
...
+stack: the typed `witness` combinator, both interpreters, and an instantiation of the
+completeness theorem. Everything here reduces by `rfl`/`decide`, so this file doubles as a
+regression test for the executable semantics.
+
+The circuit mirrors the smallest interesting PS example: witness two field elements,
+multiply them (one witnessed product + one `r1cs` constraint), and assert the result
+equals a constant.
+-/
+
+namespace Snarky.Example
```
martyall, line 18: You need to extract the r1cs stuff out of example, it blongs in some other namespace like Constraints/R1CS or something. see the purescript namespacing for inpsiration

formal/Snarky/Types.lean
kejace: Moved to a dedicated `Snarky/Vec.lean` with the full rationale in the module docstring. (To be clear, we don't reimplement `Vector` — it's targeted replacements for the core operations whose `Array.mapM`-based implementations are compiled by well-founded recursion and hence opaque to kernel reduction.)

formal/Snarky/Types.lean
kejace: Everything in the library is validated by `decide`, which reduces in the kernel — and core `Vector.map` runs through `Array.mapM`, whose loop is compiled by *well-founded recursion* and doesn't reduce in the kernel. One such call anywhere in an executable path silently wedges every downstream `decide` (this bit us: the witness combinator originally used `Vector.map` and every prover-run `decide` got stuck). `mapVec` takes the structural `List.map` route instead. Now documented at length in `Snarky/Vec.lean`.

formal/Snarky/Types.lean
kejace: Moved to the new `Snarky/DSL.lean` (the PS `Snarky.Circuit.DSL` surface). One note: `witness` isn't prover-specific — it's PS's `exists`, a circuit-side combinator that runs under *both* interpreters (the builder allocates its variables and emits its `check` constraints; only the payload is prover-only). So it lives with the user-facing combinators rather than in the prover module.

formal/Snarky/Types.lean
kejace: Moved to `Snarky/DSL.lean` alongside `witness` — it's prover-side (an `AsProver` computation) but part of the user-facing DSL surface, matching PS where `read` is exported from `Snarky.Circuit.DSL`.

formal/Snarky/Types.lean
kejace: Agreed — moved to `Snarky/DSL.lean`. `Types.lean` now holds only the `CircuitType`/`CheckedType` classes and the `FVar`/`BoolVar` instances.

formal/Snarky/Types.lean
kejace: Moved to `Snarky/DSL.lean` with the other combinators.

formal/Snarky/Example.lean
kejace: Done, following the PS namespacing: `Snarky/Constraint/Basic.lean` (the `BasicSystem` interface, PS `Snarky.Constraint.Basic`) and `Snarky/Constraint/R1CS.lean` (the R1CS backend: constraint type, instance, decidable `holds`, and its `Assignments.Le` monotonicity). `Example.lean` keeps just the `Fin 17` circuit and its `decide`-checked runs.

PR #235, opened 2026-07-14
formal: snarky DSL port, 5 — the soundness bridge to Kimchi's Generic gate

kejace:
Stacked on #234. The headline of the Kimchi bridge — and the payoff the whole `Snarky` port was pointing at.

## What it proves

Composes two results already in the stack:
- **`Snarky.prove_sound`** (#226) — a successful prover run's final assignment satisfies every constraint the builder emits;
- **`GateConstraint.holds_iff_row`** (#234) — each such constraint, evaluated, is a satisfying `Gate.Generic` row.

into the headline:

> **`satisfies_of_prove`** — if the prover run of a DSL circuit (over the Generic-gate backend) succeeds, then the constraints the builder emits translate, against the prover's final assignment, to a list of `Kimchi.Gate.Generic` rows that all `Hold`.

So the DSL builds genuine, satisfiable Kimchi Generic-gate circuits, and the prover's witness is a satisfying assignment for them. `Gate.Generic.Holds` is exactly the predicate `Kimchi.Index.rowSatisfies` dispatches to for generic rows, so this lands on the same semantics the quotient and verifier layers consume.

## Pieces

- **`toCircuit`** — a structural (not `mapM`) translation of a constraint list to a Generic gate circuit against an assignment; structural so it reduces under `decide` and inducts cleanly.
- **`toCircuit_sound`** — the list-level lift of `holds_iff_row`.
- A concrete end-to-end `decide` example (the mul circuit's constraints → a satisfied gate list).

## Scope (unchanged from #234)

Bridges to the **un-wired** `Gate.Satisfies (List Generic)`. Variable sharing holds by construction but is not enforced as a copy constraint; the wired `Kimchi.Index` bridge (domain synthesis + placement) remains future work.

## Verification

- `lake build Snarky` clean, `scripts/check-style.sh` clean
- `scripts/check_axioms.sh` green: **92 roots**, `satisfies_of_prove` and `holds_iff_row` added; closure `[propext, Quot.sound]`

### DIFF
--- a/formal/Snarky/Kimchi/Soundness.lean
+++ b/formal/Snarky/Kimchi/Soundness.lean
@@ -1,0 +1,73 @@
+import Snarky.Kimchi.Backend
+import Snarky.Laws
+
+/-!
+# The soundness bridge: a successful prover run is a satisfied Kimchi gate list
+
+The headline of the port's Kimchi bridge. `Snarky.prove_sound` says a successful prover
+run's final assignment satisfies every constraint the builder emits; `holds_iff_row`
+(from `Snarky.Kimchi.Backend`) says each such constraint, evaluated, is a satisfying
+Generic gate row. Composing them: **a DSL circuit whose prover run succeeds translates to
+a list of `Kimchi.Gate.Generic` rows that all `Hold`** (`satisfies_of_prove`) — i.e. the
+DSL builds genuine, satisfiable Kimchi Generic-gate circuits, and the prover's witness is
+a satisfying assignment for them.
+
+`Gate.Generic.Holds` is the predicate `Kimchi.Index.rowSatisfies` uses for generic rows,
+so this lands on the same semantics the rest of the Kimchi stack (quotient, verifier)
+consumes. As in `Backend.lean`, the bridge is to the **un-wired** gate list; copy
+constraints (variable sharing as wiring) are the deferred `Index` step. The executable
+demos live in `Snarky.Kimchi.Example`.
+-/
+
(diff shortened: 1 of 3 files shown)


formal/Snarky/Kimchi/Soundness.lean
```diff
@@ -0,0 +1,71 @@
...
+/-- Translate a whole constraint list into a Generic gate circuit against an assignment:
+each constraint becomes one row (`GateConstraint.toRow`), `none` if any operand fails to
+evaluate. Structural recursion (not `mapM`) so it reduces under `decide` and inducts
+cleanly. -/
+def toCircuit : List (GateConstraint F) → Assignments F → Option (List (Kimchi.Gate.Generic F))
+  | [], _ => some []
+  | con :: rest, env =>
+    match con.toRow env with
+    | some row => (toCircuit rest env).map (row :: ·)
+    | none => none
```
martyall, line 39: how can `toRow` fail?

martyall (changes requested):
I think we should hold of on this on, let's finalize the previous work first

formal/Snarky/Kimchi/Soundness.lean
kejace: Because `toCircuit`/`toRow` are stated against an *arbitrary* assignment, where a constraint can mention a variable that was never assigned. After a successful prover run it can't fire — and that's a theorem (`toCircuit_sound` produces the `some`), not a default. The docstring now says exactly this.

PR #249, opened 2026-07-17
make lean-dep-graph: vertical layout + per-package colours

martyall:
The Lean module dependency graph (`make lean-dep-graph`) goes vertical and gains per-package colour.

- **Top-to-bottom layout** (`rankdir=TB`): an edge `A -> B` (module A imports B) points **downward**, so an arrow pointing down reads "depends on" and dependencies sink toward the bottom.
- **Per-package colours**: each package cluster is filled with its own tint — pasta blue, poseidon green, bulletproof-pcs orange, kimchi purple, snarky red — with white module nodes for readable text.
- The target now also **renders `formal/docs/module-deps.svg`** via graphviz (was `.dot`-only), and the committed SVG reflects the new style.

Scope: `Makefile` (dep-graph target only), `formal/scripts/module-deps.sh`, and the regenerated `formal/docs/module-deps.{dot,svg}`.


martyall:
Folded into #248 (bulletproof-encapsulation) instead of a standalone branch.

PR #257, opened 2026-07-20
formal: box the kimchi 2nd-level namespaces in the module graph

martyall:
Draw the kimchi package cluster with nested subclusters for each **2nd-level namespace** — `Gate`, `Index`, `Permutation`, `Protocol`, `Verifier` — instead of one flat box.

Grouping rule: a module joins a box when it is at least three segments deep, so its first two segments name the box. That keeps `Gate.Semantics.*` inside the `Gate` box rather than giving it a box of its own (2nd level only, never deeper). Shallow modules — `Domain`, `Lift`, `Aggregate`, `SchwartzZippel`, `Shifted`, `GrandProduct` — float free in the cluster. Only the kimchi package is grouped; the other packages stay flat.

Pure tooling/artifact change (`scripts/module-deps.sh` + the regenerated `docs/module-deps.{dot,svg}`); no library sources touched.


martyall:
Folded into #259 (single PR against main).

PR #260, opened 2026-07-22
formal/kimchi: production chunking — the general verifier, terminal soundness roots, and the typed wire boundary

martyall:
## Summary

Generalizes the kimchi formalization from the pinned `2^σ.k = n` (one-chunk) regime to **production chunking** (`n = nc · max_poly_size`, `chunk_size = 2^(domainLog2 − σ.k)`), exactly as proof-systems implements it — and ends with `nc = 1` hardcoded **nowhere**: it survives only as the one-chunk case of the general statements.

## What's in here

**Fixtures and latent bugs.** New production `nc = 2` proof fixtures on both curves (`n = 32`, `max_poly_size = 16`; recorded by `tools/fixture-dump`'s `kimchi_proof_dump_nc2`, accepted by the production Rust verifier). Bringing them up exposed three latent bugs invisible at `zk_rows = 3` / Vesta-only, all fixed:
- `zkpm` is production's 3-factor `(X−ω^{n−zk})(X−ω^{n−zk+1})(X−ω^{n−1})`, not the full vanishing window;
- the Poseidon MDS is per-curve (`G::sponge_params().mds`) — now a gate parameter pinned by the key–index correspondence;
- production **zeroes the σ columns on the interior mask rows** `[n−zk+2, n−1)` (constraints.rs:538–544) — the committed-column model, `Nondegenerate`, and copy soundness now reflect this.

**The chunked protocol reduction.** The batch becomes the 44-row chunked assembly — the **public row joins the batch**: at `nc > 1` the public evaluations are proof-carried adversarial data, tied to the public input only through the batched opening against the verifier-computed public commitment (pinned via the key's new Lagrange chunk-commitment correspondence). The scalar side reads chunk-combined claims (`evals.combine` at `ζ^max` / `(ζω)^max`); extraction consumes `Bulletproof.chunked_batch_soundness` directly; the ft row is the double `ζ^{2^σ.k}` collapse of both the `f_comm` and `t_comm` sides.

**Terminal roots.** `Kimchi.Verifier.kimchi{Vesta,Pallas}_run_sound_algebraic_ft`: from a genuine acceptance at production chunking, AGM representations of the run's `44·nc + 1` flat segment rows and the `t` chunks deliver the guarded `Satisfies idx (pubView idx pub) wTab`. The two roots are thin wrappers over one **curve-generic terminal proof** taking the smul collapse and the Fiat–Shamir tree as hypotheses — each root applies exactly one `kimchi_fiat_shamir_*` instance. Notably there is **no `ζⁿ ≠ 1` guard** anymore — the public claims are bound batch data, so the `nc = 1` barycentric reconciliation (and its degeneracy guard) has no analogue.

**Deletion sweep.** The entire `nc = 1` verifier/reflection/reduction/capstone stack is deleted; the chunked layer holds the plain names.

**The typed wire boundary.** The final design (worked out in review): *type-level facts are exactly what serde enforces; runtime guards are exactly what verifier.rs checks.*
- `Kimchi.Verifier.Wire` is **a parser, nothing more**: serde-typed wire records (fixed `[T; N]` dims as `Vector`s, chunk payloads as unchecked `Array`s) and `Wire.{KimchiProof,KimchiVK}.check` — the verifier's own length checks factored as a total parse into the checked records. A `KimchiProof C nc` cannot hold a ragged proof; the parse is the proof. The two deliberate deviations from that rule are declared in the module docstrings: the `w_comm`/`z_comm` chunk pinning (a modeling strengthening — production feeds ragged commitments into the batch equations rather than `Err`-ing) and the `public`-count guard substitution (bounds guards in place of `verifier_index.public` equality, recovered at the soundness layer via `pubView`).
- The checked records own the plain names and the whole soundness story; **`Kimchi.Verifier.kimchiVerify` (over checked records) is the verifier**. Clients — the fixture driver, any host — compose check-then-verify themselves.
- **No file in the protocol, reduction, reflection, or capstone trees imports `Wire`** — the separation is enforced by the import graph. This also dissolved the shape-extraction machinery (`shapeBad`, `RunShapes`, per-consumer size lemmas): chunk uniformity is definitional.
- Sizing is carried in types all the way down the stream layer: the capstone's `streamPos` is `Fin`-valued (no `getElem!` reads anywhere on the reflection path), and `Ipa.transcript` returns its round challenges as a `Vector` sized by construction.

## Trust surface

The axiom gate audits 48 roots (the 45 theorem roots plus the executable path — `kimchiVerify` and the two `Wire.check` parses — so `ofReduceBool` cannot enter the deployed path unaudited). Everything reduces to: `propext, Classical.choice, Quot.sound, Lean.ofReduceBool` + `Bulletproof.poseidon_fiat_shamir_{vesta,pallas}` + `Kimchi.Verifier.kimchi_fiat_shamir_{vesta,pallas}` (the deployed-run Fiat–Shamir pair, stated over checked data at the flat segment transcript, with verifier.rs citations for the idealized absorb/squeeze schedule). DL-binding stays a hypothesis — the terminal docstrings now name the full hypothesis surface (the AGM representations, `hbind` with its standard-model caveat, the good-combination-challenge guards); the key–index correspondence (`KimchiVK.Corresponds`, now including the Lagrange chunk pin) is adjudicated numerically by `check_vk_correspond`.

## Validation

All CI-wired and green: build, style, the axiom gates, and the fixture drivers.

- **Accept**: the composed verifier accepts the production proofs — one-chunk plus `nc = 2` on both curves.
- **Reject (corruptions)**: z-eval at ζ chunks 0 **and 1**, the ζω side, `w[0]` chunk 1, `t_comm` chunks 0 **and 13** (the second collapse group, new at `nc = 2`), `ft_eval1`, and the carried public eval — on both `nc = 2` curves.
- **Reject (parse)**: `Wire.check` returns `none` on ragged eval chunks, ragged VK σ chunks, oversized `t_comm` (`> 7·nc`), a popped `lr` pair, and missing `evals_public` at `nc > 1` (production's `MissingPublicInputEvaluation`).
- **VK↔index correspondence** at both regimes (one-chunk and `nc = 2`), with the σ columns built from the model's own `Index.build?`/`sigmaAddrRow` — a drift in the interior-mask zeroing now fails CI rather than hiding behind a script-local re-implementation.
- Index, linearization, and permutation checks; the shape-literal gate (now scanning the fixture drivers too, with a scanned-file floor); cross-package deadcode reachability.

## Review response (second round)

A four-axis local review (exposition / static typing / doc-comment rigor / fixture exhaustiveness) was enacted in `7615508a..78cf51d1`:

- `7615508a` — trust-surface prose and ledger exactness: single-pair (not "grid") soundness docstrings, `Nondegenerate`'s masked-row collapse, dangling lemma references, roots.txt axiom-footprint comments, the declared wire-boundary deviations, `coeffCols`/`wCols` role consistency.
- `0c1af4cb` — the curve-generic terminal capstone and `Fin`-typed `streamPos` (deleting the bang-read shadow-lemma layer), plus the DL-binding/AGM trust story in the terminal docstrings.
- `2eeafa35` — `Vector`-sized IPA round challenges shared between the cold and warm verifiers.
- `78cf51d1` — the coverage expansion above (restored nc = 1 correspondence, parse negatives, chunk-1/ζω/high-`t` corruptions, gate hardening).


formal/kimchi/Kimchi/Index/Aggregate.lean
```diff
@@ -746,15 +747,15 @@ degenerate pairs are confined to `7·(n − zkRows)` affine lines) and runs the
 headline over it. The field bound is vacuous at Pasta (`(K+1)² ≪ 2²⁵⁵`). -/
 theorem satisfies_iff_fullFamily_dvd [Fintype F] (idx : Index F n)
     (pub : Fin idx.publicCount → F) (wTab : Fin n → Fin 15 → F)
-    (hF : (7 * (n - idx.zkRows) + 1) * (7 * (n - idx.zkRows) + 1) ≤ Fintype.card F) :
+    (hF : (7 * n + 1) * (7 * n + 1) ≤ Fintype.card F) :
```
martyall, line 759: can we justify this?

formal/kimchi/Kimchi/Protocol/Linearization.lean
```diff
@@ -84,15 +85,17 @@ def permScalar (β γ α zkpmZ : F) (e : Evals F) : F :=
     * ∏ i : Fin 6, (γ + β * e.s i + e.w ⟨i, by omega⟩))
 
 /-- The permutation vanishing polynomial at a point:
-`zkpm(ζ) = ∏ⱼ (ζ − ω^j)` over the `zkRows` masked rows `j ∈ [n − zkRows, n)`. -/
+`zkpm(ζ) = (ζ − ω^{n−zkRows})(ζ − ω^{n−zkRows+1})(ζ − ω^{n−1})` — production's
+three-factor `permutation_vanishing_polynomial` (permutation.rs:105–121), which
+coincides with the full `∏_{[n−zkRows, n)}` window only at `zkRows = 3`. -/
 def zkpmEval (n zkRows : ℕ) (ω ζ : F) : F :=
-  ∏ j ∈ Finset.Ico (n - zkRows) n, (ζ - ω ^ j)
+  (ζ - ω ^ (n - zkRows)) * (ζ - ω ^ (n - zkRows + 1)) * (ζ - ω ^ (n - 1))
```
martyall, line 92: this looks strange, is this actually correct?

formal/kimchi/Kimchi/Verifier/Kimchi.lean
```diff
@@ -64,110 +54,100 @@ structure PointEvaluations (F : Type*) where
...
+
+/-! ## The checked records -/
+
+/-- The public-evaluation source, resolving production's control flow
+(verifier.rs:332–379): carried evaluations are accepted at any `nc` and REQUIRED at
+`nc > 1`; the barycentric fallback exists only at `nc = 1` (it needs `ζ`, so it is
+computed in the verifier body). -/
+inductive PubEvalSrc (C : Ipa.CommitmentCurve) (nc : ℕ) where
+  | carried (pe : PointEvaluations (Vector C.ScalarField nc))
+  | barycentric (h : nc = 1)
```
martyall, line 96: so even the rust code distinquishes this, or it's just possible to do for `nc=1`

formal/kimchi/Kimchi/Verifier/Capstone/Standard.lean
```diff
@@ -142,7 +41,7 @@ private theorem combinedCommitment_reindex {F G : Type*} [Field F] [AddCommGroup
   rw [hC i]
 
 /-- `combinedInnerProduct` congruence across a first-index-count equality. -/
-private theorem combinedInnerProduct_reindex {F : Type*} [Field F] (ξ r : F)
+theorem combinedInnerProduct_reindex {F : Type*} [Field F] (ξ r : F)
```
martyall: can be private?

formal/kimchi/Kimchi/Verifier/Capstone/Standard.lean
```diff
@@ -1,138 +1,37 @@
...
+  Kimchi.Protocol.Equation CompElliptic.Fields.Pasta Kimchi.Verifier
 
-/-! ## The per-node Fiat–Shamir transport -/
+/-! ## Reindexing congruences for the batched combiners -/
 
-/-- `combinedCommitment` congruence across an index-count equality: reindexing the
-commitment family along `Fin.cast` preserves the polyscale combination. -/
-private theorem combinedCommitment_reindex {F G : Type*} [Field F] [AddCommGroup G]
+/-- `combinedCommitment` congruence across an index-count equality. -/
+theorem combinedCommitment_reindex {F G : Type*} [Field F] [AddCommGroup G]
```
martyall: private?

formal/kimchi/Kimchi/Verifier/Reduction/Soundness.lean
```diff
@@ -123,63 +60,119 @@ private theorem bound_unique [Field F] [AddCommGroup G] [Module F G] (σ : SRS G
...
+
+/-- The chunked key–index correspondence: the committed chunk columns are the circuit's
+own. -/
+def VKCorresponds [Field F] [AddCommGroup G] [Module F G] (σ : SRS G) (nc : ℕ)
+    {n : ℕ} (comms : IndexComms (Fin nc → G)) (idx : Index F n) : Prop :=
+  comms = indexerOf σ nc idx
+
+/-! ## The batch assembly (44 logical rows)
+
+Row indices `0–42` keep the `nc = 1` layout (witness `0–14`, accumulator `15`, σ
```
martyall: is this what you find in the rust code? This seems sort of ad-hoc, there is no public api of this library for which you need to maintain backward compatibility, any piece can be moved or changed

formal/kimchi/scripts/check_kimchi_verifier.lean
```diff
@@ -1,50 +1,94 @@
...
-the complete wire proof that production `verify` accepts, and reject corruptions of
-an evaluation, the quotient commitment, and `ft_eval1`. This is the composed
-adjudication: every transcription judgment in `Kimchi/Verifier/Kimchi.lean` (absorb
-orders, the public-commitment blinding, the warm-sponge handoff) either reproduces
-production's accept bit or fails here. -/
+/-! The executable kimchi verifiers against production proofs (`tools/fixture-dump`'s
+`kimchi_proof_dump` / `kimchi_proof_dump_nc2`):
 
-open Lean FixtureKit Bulletproof Bulletproof.Fixture Kimchi.Fixture.Kimchi Kimchi.Verifier
+* the `nc = 1` verifier (`kimchiVerify`) must ACCEPT the one-chunk wire proof
```
martyall: there should be no difference between an nc=1 and nc > 1 verifier, there should only be one verifier.

formal/kimchi/Kimchi/Verifier/Capstone/Reflection.lean
```diff
@@ -449,68 +494,707 @@ private theorem pubDotFoldl_aux {F : Type*} [Field F] (ω pt : F) (l : List F)
...
-      = -((idx.pubPoly (pubView idx pub)).eval (runOracles C σ vk p pub).zeta) := by
-  rw [← barycentricPubEval_eq idx (pubView idx pub) hζn]
-  have hζN : runZetaN C σ vk p pub = (runOracles C σ vk p pub).zeta ^ n := by
-    unfold runZetaN
-    rw [powPow2_eq, ← hn]
+/-- The flat stream position of abstract batch row `i`, chunk `c` — the `to_batch`
+layout of the deployed stream: the public row's chunks first, the ft singleton at `nc`,
+then per logical row `nc` consecutive chunks (`z`, the six selectors, `w`,
+coefficients, `σ[0..6)`). -/
+def streamPos (nc : ℕ) (i : Fin 44) (c : ℕ) : ℕ :=
```
martyall: this looks pretty ad-hoc, is this what you find in the rust code?

martyall (commented):
Few comments:

1. It seems like the poseidon related changes (abstracting mds) should be its own PR, it doesn't even seem related to chunking
2. You have many instances of public bindings which could be private
3. some of the stream position / general indexing in this PR seems ad-hoc, it needs to be double checked against rust impl.
4. the doc comments do not reflect a research paper, there are scratch thoughts and comments explaining the evolution of hte codebase

martyall:
Response to the review, categorized:

### Top-level

1. **MDS abstraction should be its own PR — AGREE** (to the split). One clarification on "not related": it was the nc=2 *Pallas* fixture that exposed it — `Gate.Poseidon` hardcoded the `fp_kimchi` MDS as generic integer literals, which is garbage over `Fq`, so the Pallas terminals described a diverging verifier. It's a prerequisite of the Pallas half of this PR, but it is self-contained and I'll split it out as a standalone PR this one rebases onto.

2. **Public bindings that could be private — AGREE.** The file moves during the deletion sweep promoted several things that have no external consumers (`KimchiBatchAcc`, `combinedInnerProduct_reindex`, `selComm`/`selGate`/`evalsExt` where cross-file use has since collapsed, …). I'll do the same audit Wire got: every public binding justified by an external use or a root, else private.

3. **Stream position / indexing ad-hoc, double-check against Rust — AGREE** with the action. Two halves to it: (i) the *flat stream itself* (what the verifier feeds the IPA) is behaviorally pinned — a wrong `to_batch` order mis-combines the polyscale walk and the production fixtures would reject, so acceptance on nc=1+nc=2 both curves is a by-value adjudication of the order; (ii) but the *abstract* 44-row order and the `streamPos` bridge between them are our invention (see 6 below) — I'll make the abstract batch mirror `to_batch` order directly, which shrinks or eliminates the index bridge, and document each row against verifier.rs with line references.

4. **Doc comments read as evolution notes, not a finished document — AGREE.** "Keeps the nc=1 layout", "NEW against the nc=1 correspondence", etc. are changelog prose that belongs in commit messages. Full docstring sweep to timeless present-tense statements of what the code is.

### Inline

5. **`Aggregate.lean` — `hF` widened to `(7n+1)² ≤ |F|`: "can we justify this?" — AGREE that the justification must be written down (it's missing, and the surrounding docstring is stale — it still says "confined to `7·(n − zkRows)` affine lines").** The change itself is forced: with production's 3-factor `zkpm`, the permutation recurrence is *live* on the interior mask rows `[n−zk+2, n−1)` (the polynomial no longer vanishes there), so `Nondegenerate` — and hence the degenerate-pair counting behind `exists_nondegenerate_grid` — must range over all `n` rows, not `n − zkRows`. Still vacuous at Pasta. I'll fix the docstring to carry this argument.

6. **`Linearization.lean` — 3-factor `zkpmEval`: "is this actually correct?" — DISAGREE (no change needed).** It's a verbatim transcription of production's `permutation_vanishing_polynomial` (permutation.rs:105–121), which is three factors *regardless of* `zk_rows`; it coincides with the full window only at `zk_rows = 3`. And it is numerically adjudicated where it matters: the nc=2 fixtures have `zk_rows = 5`, where the two forms differ — with the full-window form, `permScalar`/`ftEval0` come out wrong and the verifier rejects the production proofs. This was in fact one of the three latent bugs the chunking fixtures caught (the old code had the full window).

7. **`Verifier/Kimchi.lean` — carried vs. barycentric public evals: "does Rust distinguish this?" — Yes, exactly (no change needed).** verifier.rs:332–379: if `evals.public` is `Some`, it is used at *any* chunk count; if `None` and `chunk_size > 1`, `Err(MissingPublicInputEvaluation)`; if `None` and `chunk_size = 1`, the barycentric computation. `PubEvalSrc.carried / .barycentric (h : nc = 1)` is that control flow with the `nc = 1` fact carried on the fallback constructor.

8. **`Capstone/Standard.lean` — `combinedInnerProduct_reindex`, `KimchiBatchAcc` private — AGREE** (no external consumers; part of item 2).

9. **`Reduction/Soundness.lean` — "row indices keep the nc=1 layout … seems ad-hoc, no backward compatibility to maintain" — AGREE.** The abstract row order was inertia from the deleted layer, and you're right that nothing constrains it (the library has no committed API). I'll reorder the abstract batch to mirror the deployed `to_batch` order, which is also what makes item 3 clean.

10. **`check_kimchi_verifier.lean` — "there should only be one verifier" — AGREE; the code already complies, the docstring doesn't.** As of the wire-boundary commit there is exactly one verifier (`kimchiVerify` over checked records; the script's `verifyWire` is the parse-then-verify composition, run identically on all three fixtures). The module docstring still describes the transitional two-verifier adjudication — stale, will rewrite (an instance of item 4).

11. **`Capstone/Reflection.lean` — "looks ad-hoc, is this in the Rust?" — same resolution as 3/9.** The fold→sum auxiliary lemmas there are internal proof bridges (executable fold ⇒ `Finset.sum`), not transcriptions; the Rust-facing content (the `combineAt` fold = `eval_polynomial`, the stream order) is the part that gets aligned to `to_batch` and documented with line references.

Plan of record, in order: (1) split the MDS fix into a prerequisite PR; (2) reorder the abstract batch to `to_batch` order + document indexing against verifier.rs; (3) privatization audit; (4) docstring sweep (including the `hF` justification and the driver header).

martyall:
Review items addressed (three commits pushed, plus one prerequisite PR):

1. **MDS split out as a prerequisite** — #261 (`poseidon-mds`, off main) carries the per-curve Poseidon MDS parameterization alone, ported against the nc=1 files. Once it merges, main will be merged back into this branch and the MDS hunks drop out of this diff.
2. **`eee3305f`** — the abstract 44-row batch is now in `to_batch` order (public, z, selectors, witness, coefficients, sigma), with a row table in the `Reduction/Soundness.lean` preamble citing the exact `verifier.rs` pushes. `streamPos` collapses to the uniform two-case form; the ad-hoc index bridge is gone.
3. **`cb74d834`** — privatization audit of `Kimchi/Verifier/**`: 14 helpers privatized (incl. `streamPos`, the reindex congruences, the sponge digest internals). `KimchiBatchAcc`/`nodeInput` and the checked `Corresponds` stay public because they appear in rooted theorem statements.
4. **`99a7c02b`** — docstring sweep: the `hF : (7n+1)²` bound is now justified in the `satisfies_iff_fullFamily_dvd` docstring (three-factor mask ⇒ recurrence live on interior mask rows ⇒ all-row counting), the `check_kimchi_verifier` header describes the one-verifier composition, and remaining development-history phrasing is rewritten as timeless description.

The two disagree-with-evidence items (three-factor `zkpmEval`, `PubEvalSrc` carried/barycentric) stand as argued in the earlier response.

All gates green after each commit: build, style, 45-root axiom gate (unchanged allowed set), deadcode, five fixture drivers.

PR #265, opened 2026-07-23
formal/kimchi: fixture-coverage follow-ups from the #260 review

martyall:
## Summary

Enacts the fixture-coverage findings from the #260 review (group A of the follow-up ledger): the gaps where a feature the chunking PR introduced was exercised by no fixture. All new fixtures are recorded from the production Rust prover/verifier (seeded, deterministic) — none hand-forged.

### A1 + A2 — Pallas correspondence and a zk_rows=5 model pin

`index_dump` becomes a curve-generic macro over `mixed_index_over`, emitting:
- `index_vesta.json` — unchunked regression, zk_rows=3 (**byte-identical** to before).
- `index_{vesta,pallas}_nc2.json` — chunked (`override_srs_size = n/2`), **zk_rows=5**, so the σ interior-mask range `[n−zk+2, n−1)` is nonempty (rows 29, 30 at n=32).

`check_index_fixture` is made curve-generic and loops all three, **directly** pinning `sigmaAddrRow`'s zeroing branch row-by-row at zk_rows=5 — a sign/off-by-one drift now fails a model-level check, not only the indirect commitment check. `check_vk_correspond` adds the **Pallas nc=2** correspondence (28 columns × 2 chunks); Pallas was previously untested at any nc.

### A3 — carried public evals at nc=1 (producible, dumped properly)

Finding: `ProverProof::create` **always** populates `evals.public` (prover.rs:1048), even at nc=1; the verifier's first branch consumes carried evals at any nc (verifier.rs:332). The nc=1-*without*-public case is a wire-serialization artifact (o1js / OCaml `to_repr` drops the chunks, and the verifier then recomputes barycentrically). So carried-at-nc=1 is production-reachable — `kimchi_proof_dump` now emits the one-chunk proof twice (`kimchi_proof_vesta.json` without, `kimchi_proof_vesta_pub.json` with `evals_public`, same Rust verify asserted). The driver checks the carried proof accepts via `PubEvalSrc.carried` and that a corrupted carried eval flips the verdict — proving the branch feeds verification. No forged fixture.

### A4 — parameter diversity (nc=8 landed; two items deferred)

`kimchi_proof_dump_nc8` dumps an **nc=8** proof (domain n=64, `max_poly_size=8`, zk_rows=19, 56-chunk quotient) with acceptance + a high-chunk (chunk 55) corruption. This also covers `max_poly_size ≠ n/2`. The nc=8 corruption matrix is bounded by a documented `heavy` flag (each verify is a 56-chunk batch MSM in the interpreter) that logs which corruption *kinds* it skips — all of which stay covered at nc ≤ 2.

- **nc=3 is unproducible** — a non-power-of-two `max_poly_size` misaligns the segment chunking and the real prover rejects it (`WrongBlinders`); a power-of-two domain never divides by 3.
- **public_count > 1** and **t_comm < 7·nc** are **deferred**: the mixed circuit has one public row and a full 7·nc quotient; both need circuit rework rather than a dumper flag.

### A7 — CI build guard

A new `build-fixture-dump` job in `test.yml` builds the standalone crate under its pinned **1.92** toolchain (the root pins 1.88, too old) with `submodules: recursive`, so the crate — doubled in surface by the chunking work — can no longer bit-rot unseen. Build-only (running needs the SRS). README brought current (8 binaries, per-package output map including the three indices, the two nc=1 serializations, and nc=8).

## Verifier test matrix (`check_kimchi_verifier`)

| fixture | curve | nc | accept | corruptions → reject | parse-rejects → none |
|---|---|---|---|---|---|
| vesta | V | 1 | ✓ | z0, t0, ft_eval1 | 4 |
| vesta_pub | V | 1 | ✓ | z0, t0, ft_eval1, **pubeval0** | 4 |
| vesta_nc2 | V | 2 | ✓ | t13, z0, t0, ft1, z1, zω0, w0·1, pub0 | 5 |
| pallas_nc2 | P | 2 | ✓ | (same 8) | 5 |
| vesta_nc8 | V | 8 | ✓ | **t55** (bounded, logged) | 5 |

## Validation

Independently re-run after assembly: `cargo fmt --check` + `cargo build` (1.92) green, fmt/clippy-clean on the new code; **determinism** confirmed — all fixtures regenerate byte-identically and `index_vesta.json` + the existing proof fixtures show no git change; `lake build` green; style + shape gate clean (new Lean uses the named shape constants); every fixture driver green (three indices incl. the σ-zeroing rows, VK-correspondence incl. Pallas, the five-fixture verifier matrix, linearization, permutation); axiom gate unchanged (48 roots, allowed set intact — no checked defs added). All Rust prove+verify assertions pass at dump time via the cached SRS.

https://claude.ai/code/session_01PSVEv2RHLtK4WDc4g3zZz7


.github/workflows/lean.yml
```diff
@@ -160,8 +160,12 @@ jobs:
         KIMCHI_FIXTURES_DIR=kimchi/fixtures \
           lake env lean kimchi/scripts/check_perm_fixture.lean
 
-    # The index model must accept the mixed-gate production circuit: construction
-    # by decision, derived columns, satisfiability, and corruption rejections.
+    # The index model must accept the mixed-gate production circuit at BOTH chunking
```
martyall: are these comments really necessary? They don't enforce or disambiguate anything, they just sttae the obvious of what you would see if you lookied into the script

.github/workflows/lean.yml
```diff
@@ -180,27 +184,30 @@ jobs:
 
     # The composed executable kimchi verifier (check-then-verify, both sponge
     # schedules, the scalar side, the to_batch segment stream, the warm-started IPA
-    # check) must accept complete production wire proofs — the one-chunk fixture and
-    # the nc = 2 fixtures on BOTH curves — reject a corruption matrix (both
-    # evaluation points, chunks beyond 0, both t_comm collapse groups), and refuse
-    # to parse ragged or mis-pinned wire records
-    # (fixtures/kimchi_proof_{vesta,vesta_nc2,pallas_nc2}.json, from
-    # tools/fixture-dump's kimchi_proof_dump binaries).
+    # check) must accept complete production wire proofs across nc ∈ {1, 2, 8}, both
```
martyall: are these comments really necessary? They don't enforce or disambiguate anything, they just sttae the obvious of what you would see if you lookied into the script

.github/workflows/lean.yml
```diff
@@ -180,27 +184,30 @@ jobs:
...
     # The verifier-key <-> index correspondence (VKCorresponds at production
-    # chunking), by value, at BOTH chunking regimes on Vesta — the nc = 1 VK
-    # (zk_rows = 3) and the nc = 2 VK (zk_rows = 5); Pallas has no index fixture:
-    # every committed column chunk of each production VK equals the value-MSM of
-    # that key's Lagrange-basis chunk commitments against its derived column. The
-    # sigma columns are the MODEL's own derivation (Index.build? at each regime's
-    # zk_rows, then Index.sigmaAddrRow), so the model's sigma-zeroing branch is
-    # adjudicated against production at zk_rows = 5 and the un-zeroed addresses at
-    # zk_rows = 3; selectors carry the per-chunk fixed blinder.
+    # chunking), by value, across both curves and both regimes — the Vesta nc = 1 VK
```
martyall: are these comments really necessary? They don't enforce or disambiguate anything, they just sttae the obvious of what you would see if you lookied into the script

PR #294, opened 2026-08-14
snarky-kimchi: the gadget laws apply the gate-semantics theorems directly

martyall:
Review follow-on from #293: the gadget layer's laws should apply the `Kimchi/Gate/Semantics/*` soundness and completeness theorems directly, as the Poseidon laws already do. Two spots fell short; both closed.

## `addFast_complete_spec` promises the group sum

The completeness post said only that the honest run accepts and the outputs are *readable* — weaker than the gate's own theorems, and weaker than Poseidon's complete law (which characterizes the output as `blockCipher`). It now delivers the group-level characterization: for pinned input reads and nonsingularity witnesses, *either* the infinity flag reads `1` and `P₁ + P₂ = 0` in Mathlib's group, *or* it reads `0` and the output coordinates are a nonsingular point equal to the sum.

The derivation applies the gate's `sound` to the accepted run: the row the honest prover filled satisfies the verified gate, so `Kimchi.Gate.AddComplete.sound` characterizes what was computed. The internal tail spec's post carries the honest row's concrete values (instead of bare `isOk`s) to make that application possible.

This is the fact the EndoMul slice composes for its initial accumulator `P₀ = [2](T + φ(T))`.

## EndoScalar soundness reroutes through the chain lemmas

The sound law's conclusions were already in gate-semantics vocabulary, but the proof re-derived the chain folds at the gadget carrier (`threaded_sound`'s fold induction) instead of consuming `chain_decompose`. Now:

- `chainCrumbs`, `chainCrumbs_length`, `chain_decompose` go public in kimchi (arrival-with-consumer);
- `threaded_chain` extracts the threading's structural facts — round count, seed wiring, shared accumulator variables between adjacent rounds, the final triple — valuation-free;
- `threaded_sound` becomes wiring only: the extraction instantiates `chain_decompose`'s indexed run at the payload reads. The fold arithmetic lives solely in the gate-semantics layer, and the `decomposeA_eq_table` conversions vanish from the sound proof (`chain_decompose` concludes in `decomposeA` vocabulary directly).

The `threaded_chain` bridge is the shape the EndoMul slice reuses to consume `pallas_endoMul`/`vesta_endoMul` wholesale.

## Completeness goes constructive — one source for both directions

Review of the completeness side found the same disease in every remaining proof: existentially packaged gate completeness can never certify fixed prover code, so the gadget layer was replaying the gates' algebra privately. Three commits close it:

- **AddComplete**: the gate gains its canonical row `build` and `complete_build : Holds (build …)` (the gadget's former private `valueWitness`/`valueWitness_holds`, relocated); the `∃`-shaped `complete` becomes a one-line corollary and `complete_noninf`/`complete_inf` retire. `addFast_complete_spec` consumes `complete_build` for acceptance and the gate's `sound` for the output characterization.
- **Poseidon**: `chainHolds_of_succ`'s fifteen-way constraint case analysis becomes one identification — a window whose successor is the round image *is* the gate's `build`, so the gate's `complete` applies per window.
- **EndoScalar**: the row witness computes the gate's `build`; the completeness loop's invariant identifies the prover's run with the gate's threaded `chainBuild` (uniform — round `k`'s inputs are round `k−1`'s outputs); acceptance is `chain_complete`, the final reading is `chain_decompose` — the same theorem the sound law consumes. The bare-table detour (`buildTable`/`buildTable_eq_build`/`complete_table`) and the gadget's fold bookkeeping die, and the characteristic hypotheses `h2`/`h3` drop from both completeness specs (completeness needs no field non-degeneracy — they were an artifact of the detour).

Net: every soundness and completeness fact for every landed gadget now derives from its gate's single semantics development — universal theorems applied to the accepted run, constructive completeness at the canonical witness — with the gadget layer contributing wiring only.

Gates green: build, corpus 23/23, style, snarky + kimchi axiom gates, deadcode 0, per-root lint, shake.

### DIFF
--- a/formal/kimchi/Kimchi/Gate/AddComplete.lean
+++ b/formal/kimchi/Kimchi/Gate/AddComplete.lean
@@ -29,9 +29,10 @@ carries the constraint model; the theorems below are proved in
   sum `(x₁,y₁) + (x₂,y₂)` is the group element the gate encodes — `0` when `inf = 1`,
   else the affine output `(x₃, y₃)` — using that `inf` is boolean (`inf_boolean`). It
   splits into the per-case `sound_point_noninf` / `sound_point_inf`.
-* `complete` — COMPLETENESS, both cases in one statement: for on-curve inputs (`y₁ ≠ 0`),
-  an honest prover can fill a satisfying witness, casing internally on whether the sum is
-  finite or `∞`. Splits into the per-case `complete_noninf` / `complete_inf`.
+* `build` / `complete_build` — COMPLETENESS, constructive: the canonical row the honest
+  prover fills, and the theorem that it satisfies the gate for on-curve inputs
+  (`y₁ ≠ 0`), casing internally on whether the sum is finite or `∞`. `complete` is the
+  existential corollary.
 -/
 
 namespace Kimchi.Gate.AddComplete
--- a/formal/kimchi/Kimchi/Gate/Semantics/AddComplete.lean
+++ b/formal/kimchi/Kimchi/Gate/Semantics/AddComplete.lean
@@ -73,122 +73,93 @@ theorem sound_noninf
       WeierstrassCurve.Affine.negAddY, WeierstrassCurve.Affine.addX, ha1, ha2, ha3]
     linear_combination -c5 - w.s * c4
 
-/-- COMPLETENESS: the honest prover can always fill in the witness. For on-curve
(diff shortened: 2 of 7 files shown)


martyall:
Superseded by the per-gadget split, each branching off main so they can be reviewed and merged independently: #295 (AddComplete), #296 (Poseidon), #297 (EndoScalar). Content identical to this branch's four commits, regrouped by gadget.

PR #298, opened 2026-08-15
snarky-kimchi: the endoMul gadget — [s]·T with s the EndoScalar decode, sound and complete

martyall:
The EndoMul slice, five commits: the gadget, its constraint semantics, and the law family proving the defining equation in both directions — the gadget's output is `[s]·T` with `(s : F) = EndoScalar.toField crumbs λ` and the scalar as the crumbs' `nReconstruct`. EndoMul multiplies by exactly the scalar EndoScalar decodes, one shared crumb list, both directions from the gate-semantics source.

## The gadget (d44a1790)

`Snarky.Kimchi.endoMul eb rounds g scalar`, the port of PS `endo` (OCaml `Pickles.Step_main_inputs.Ops.endo`): the bulk MSB-first bit witness, the sealed `β·x` and two `addFast .checkFinite` building `P₀ = [2](g + φ(g))`, one eight-variable witness per GLV round via the gate's own `build`, the scalar pin, one `endoMul` constraint. Corpus 24/24 including `endo_mul_step_circuit` with raw per-cell variable ids.

## The semantics (b6d8ebdd)

`.endoMul` reads as a SUCCESSOR CHAIN: each round is the verified gate at the payload's endo coefficient, its output cells taken from the next round's `p`/`nAcc` values, the last round's from the payload finals — the two-row gate's next-row read, value-level. The payload carries `endo : F` as parameter data (reduce ignores it — corpus-neutral), and the gate gains `ok`/`ok_iff`. The successor-read design makes the laws' row threading definitional: a round's output cells and its successor's input cells are the same variables.

## Soundness (1ba44e5d)

Kimchi factoring first: `endoMul_off` extracts the capstone composed with `accumulator_chain`, generic over the off-targets fact — `pallas/vesta_endoMul` become two-line instantiations; `chain_nAcc` reads the final register as `nReconstruct` of the run's crumbs; `crumbList` is publicized with validity/length lemmas; the VarBaseMul toolkit gains `y_ne_zero_of_odd_order`. Then `endoMul_spec` (generic over field + off + eigenvalue) with deployed corollaries `endoMul_spec_pallas/vesta`. The gadget layer contributes wiring only: the cons-shaped threaded extraction, self-reads making the threading rfl, and the init chain through `sealVar_spec` + the new `addFast_checkFinite_spec` (the pinned-mode soundness form; `addFastTail_spec` now records the returned flag).

## The produce chain (2def8078)

`chainBuild` threads the gate's canonical row from the init accumulator, and `chain_complete` (+ `pallas/vesta_chain_complete`) certifies every row of the honest walk, generic over off. The non-degeneracy is PRODUCED forward rather than read off an accepted run: `window_produce_core` drives `secant_add` twice through `stepWindow`'s own formulas — an affine point is never zero, and the negation cases are priced by `combo_ne_zero` (a positive bounded two-base combination shifted into off's forbidden set).

## Completeness (440ae754)

`endoMul_complete_spec_pallas/vesta`, deployed-only as the preconditions collapse only at Pasta: on a readable in-range faithful scalar and a readable on-curve base, the honest run accepts and the returned point reads as `[s]·T` at the scalar's canonical crumbs. The loop invariant identifies the run with `chainBuild` and carries the successor chain's check with the current row's inputs as its finals; per-round acceptance is the produce chain's; the register pin is `chain_nAcc` through the new bit-to-crumb bridge `crumbList_ofBits` (the testBit-MSB stream IS `crumbsOf`); the point grant is `endoMul_off` at the walk.

## Trust-surface notes for review

- The snarky axiom gate gains `deployedRoots`: the four deployed-at-Pasta laws may carry the certified upstream `native_decide` witnesses, discriminated by defining module exactly as kimchi's gate does; all 117 other roots stay pure core Lean.
- Publicized (each already load-bearing in a public statement or needed by the law layer): kimchi `crumbList`, `chainBuild_eta`; pasta `pallas/vesta_endo_nonsingular`.
- The #293 promotion watch resolves as unneeded: `rowWit` reads single CVars, so `mapM_eval_ok`/`readAll_ok` gain no second consumer.

Full battery green per commit: build, CS-equality corpus 24/24, style, kimchi (106) + snarky (121) + pasta (13) axiom gates, deadcode 0, env linters, shake.

### DIFF
--- a/formal/kimchi/Kimchi/Gate/EndoMul.lean
+++ b/formal/kimchi/Kimchi/Gate/EndoMul.lean
@@ -170,6 +170,14 @@ instance [DecidableEq F] (endo : F) (w : Witness F) : Decidable (Holds endo w) :
   unfold Holds
   infer_instance
 
+/-- EXECUTABLE checker — runnable on a concrete witness. -/
+def ok (endo : F) (w : Witness F) : Bool :=
+  (constraints endo w).all (· == 0)
+
+/-- Reflection: the checker faithfully decides the relational constraints. -/
+theorem ok_iff (endo : F) (w : Witness F) : ok endo w = true ↔ Holds endo w := by
+  simp only [ok, Holds, List.all_eq_true, beq_iff_eq]
+
 omit [DecidableEq F] in
 /-- `Holds` as the readable 12-conjunction (what the soundness proofs destructure). -/
 theorem holds_iff (endo : F) (w : Witness F) :
--- a/formal/kimchi/Kimchi/Gate/Semantics/EndoMul.lean
+++ b/formal/kimchi/Kimchi/Gate/Semantics/EndoMul.lean
@@ -523,6 +523,83 @@ theorem crumbList_valid (endo : F) (m : ℕ) (g : ℕ → Witness F)
   · rcases hb3 with h3 | h3 <;> rcases hb4 with h4 | h4 <;> rw [h3, h4] <;> norm_num
 
 omit [DecidableEq F] in
+/-- A natural's low base-4 digit, cast to the field, as its two bits — the shape a
+    window's second crumb `b₄ + 2·b₃` takes at bit values. -/
+private theorem cast_mod_four (k : ℕ) :
+    ((k % 4 : ℕ) : F)
+      = (if k.testBit 0 then (1 : F) else 0) + 2 * (if k.testBit 1 then 1 else 0) := by
+  have h0 : k.testBit 0 = decide (k % 2 = 1) := Nat.testBit_zero k
+  have h1 : k.testBit 1 = decide (k / 2 % 2 = 1) := by
+    rw [show (1 : ℕ) = Nat.succ 0 from rfl, Nat.testBit_succ, Nat.testBit_zero]
+  rcases Nat.mod_two_eq_zero_or_one k with he | he <;>
+    rcases Nat.mod_two_eq_zero_or_one (k / 2) with he' | he' <;>
+    rw [show k % 4 = k % 2 + 2 * (k / 2 % 2) from by omega] <;>
+    simp [h0, h1, he, he'] <;> norm_num
+
+omit [DecidableEq F] in
+/-- **The bit-to-crumb bridge.** When each row's four bits are the scalar's bits
+    MSB-first — row `r` holding bits `4m−1−4r` down to `4m−4−4r`, the prover's bulk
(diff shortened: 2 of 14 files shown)


martyall:
Amendment `1531d15b`, addressing the generic-composition concern: the completeness walk was already generic (it was the private `endoMul_complete_core`); it is now the public `endoMul_complete_spec`. The curve facts — eigenvalue, endomorphism curve-stability, GLV off-targets, the nonzero `(1+λ)`-multiple — are hypotheses, the Lean rendering of the PS typeclass dictionary, so pickles-layer completeness laws stated over a generic field (as in the PS package, down to wrap/step main) take the same dictionary and thread it to this law, exactly as this walk threads `W` and its facts into `addFast_complete_spec`. Everything discharges once at the deployed instantiation.

A pleasant consequence the axiom gate now certifies: the generic law sits in the gate's regular roots, so its closure is the three standard axioms alone — the certified `native_decide` witnesses enter only with the deployed pair.

martyall:
Amendment `ac49304b`: the dictionary is now the `HasEndo` structure — the PS class plus the ambient curve facts as one explicit bundle (curve, endo coefficient, eigenvalue, short shape, prime odd order, field characteristics, eigenvalue relation, endo curve-stability, GLV off-targets, nonzero `(1+λ)`-multiple). Both generic laws take `d : HasEndo F`; the four deployed corollaries are one-line instantiations at `HasEndo.pallas`/`HasEndo.vesta`, which absorb all the discharge terms. A structure rather than a class, since the formal tree threads theorem content explicitly; the data fields are literals so the deployed projections reduce and the corollaries' concrete statements are byte-for-byte unchanged. Composite generic circuit laws now thread a single argument down to endoMul.

martyall:
Amendment `2878d93a`: the four per-curve corollaries retire — the generic laws are concretized only inside a larger circuit's instantiation, never on their own, so per-curve statements carried no consumer. The deployed surface is now the two dictionary values `HasEndo.pallas`/`HasEndo.vesta`: the discharge artifacts a larger circuit applies the generic laws at, and (by type-checking) the exhibit that the dictionary is satisfiable at Pasta — the anti-vacuity role the corollaries used to play. The axiom gate's `deployedRoots` becomes exactly those two values, so every law theorem in the snarky package is now certified pure core Lean, with the package's entire `native_decide` trust localized in the dictionaries.

PR #323, opened 2026-09-02
pickles: the linearization, ported and proved to compute the gate contribution

martyall:
kimchi's linearization is a 4220-token reverse-Polish program. Until now this tree adjudicated it only *by value* — `check_linearization.lean` said of the stream that it "never appears in a Lean statement". This branch makes that sentence false, on three independent legs.

## What is proved

```lean
theorem evaluate_fpTokens (α β γ jc van : Fp) (e : Evals Fp) :
    evaluate (e.toEnv Pasta.pallasEndo symMds α β γ jc van … ) fpTokens
      = gateLinearization Pasta.pallasEndo symMds α e

theorem circuit_gateLinearization (inp : Inputs Fp) (α : Fp)
    (htab : ∀ n, (inp.alphaPows n).val V = α ^ n) :
    ⦃⌜True⌝⦄ evaluate (inp.toEnv …) fpTokens
    ⦃⇓ a _ => ⌜a.val V = gateLinearization Pasta.pallasEndo symMds α (inp.evals.map (·.val V))⌝⦄
```

The first says the deployed stream computes the closed form at **every** evaluation and challenge, on both sides of the Pasta cycle. The second says every valuation satisfying the constraints the in-circuit interpreter emits reads its output as that same quantity — the gate contribution to `ft_eval0`.

This is **relative** faithfulness: that the circuit computes what the wire protocol computes. Whether the protocol is sound is out of scope (`docs/soundness-line-retirement.md`).

## How

Proof by reflection. Both sides run at `CMvPolynomial 64 K`, one formal variable per environment input, and the polynomials are compared by compilation. The whole program collapses to **486 monomials** — the linearization is a sum of small per-gate constraints, not a nested product, so normalising it does not blow up, Poseidon's degree-7 S-boxes included.

Two transport lemmas carry the polynomial identity back to the field, neither of which inspects the token array, so the development costs nothing per token:

- `evaluate_map` — the machine commutes with a carrier map intertwining the environments
- `gateLinearization_map` — the closed form transports along an `F`-algebra map, nearly free because every `Argument` already carries its naturality square as `constraints_map`

`gateLinearization` is generalised in place to read at an arbitrary `F`-algebra — the freedom `Argument.constraints` always offered and nothing used. At `R := F` it elaborates exactly as before.

## Constraint-system equality

`linearization_step_circuit` joins the CS-equality corpus and passes: same 474 gates, types, coefficients, wiring and public size as the PureScript dump, which already matches OCaml. Gated on every PR.

Diagnosed by measurement rather than guesswork: 569 rows against 474, and the 62 live α-sites cost exactly 95, so the Alpha+Pow peephole was the entire divergence and nothing else about the port was wrong.

## Trust surface

`Reflect/Certificate.lean` is the only `native_decide` site in the tree outside CompElliptic and `Pasta/Endo.lean`, and `pickles/scripts/check_axioms.sh` is the gate that says so — verified by negative control.

## Not done

- **Completeness.** Without it the soundness statement is satisfiable by a circuit no assignment satisfies. `Reflect/Soundness.lean` says so rather than leaving it implied.
- **The Fq column of the CS corpus.** `fq_reflects` and `evaluate_fqTokens` are proved, but `KimchiFixture.PS.Raw` is Fp-typed and its `build` synthesises the domain from Pallas constants, so nothing checks the Fq circuit against recorded data. Same blocker `group_map_wrap` and `combine_poly_wrap` wait on.
- **The general `skipIfNot`.** `evalLoop_spec` takes the disabled-branch facts as explicit hypotheses; the specialisation is in that theorem's statement rather than the relation. Only matters if optional features are ever enabled, which needs kimchi's fragment to grow lookups first.

https://claude.ai/code/session_01RjRVAXHUPhtxTyPTqKMz3r


formal/pickles/PicklesFixture/Emit.lean
```diff
@@ -0,0 +1,124 @@
+import Pickles.Linearization.Types
```
martyall: this should just be in some scripts subdir to distinguish it from library code

formal/pickles/PicklesFixture.lean
```diff
@@ -0,0 +1,10 @@
+import PicklesFixture.TokenParser
```
martyall: this should just be in some scripts subdir to distinguish it from library code

formal/pickles/PicklesFixture/TokenParser.lean
```diff
@@ -0,0 +1,233 @@
+import Pickles.Linearization.Types
```
martyall: this should just be in some scripts subdir to distinguish it from library code

formal/pickles/Pickles/Linearization/Circuit.lean
```diff
@@ -0,0 +1,399 @@
...
+  /-- The joint combiners agree. -/
+  jointCombiner : ce.jointCombiner.val V = pe.jointCombiner
+  /-- The `β` challenges agree. -/
+  beta : ce.beta.val V = pe.beta
+  /-- The `γ` challenges agree. -/
+  gamma : ce.gamma.val V = pe.gamma
+  /-- Related branches select related results. Conditional rather than pinned to the
+  disabled branch, so the enabled case stays stateable — the feature predicate is a
+  parameter of both environments. -/
+  ifFeature : ∀ (f : FeatureFlag) (t₁ n₁ : CircuitM F (Builder V c) (FVar F)) (t₂ n₂ : F),
```
martyall: why is this pinned to the circuit builder?

formal/pickles/Pickles/Linearization/Circuit.lean
```diff
@@ -0,0 +1,399 @@
...
+same readings — provided the α-table reads as the powers of `α`, which is the caller's
+obligation and what `precomputeAlphaPowers` discharges. -/
+theorem inputs_circuitCompatible [LawfulBasicSystem F c] {V : Valuation F} (endo : F)
+    (mds : Kimchi.Gate.Poseidon.Mds F)
+    (lk : Kimchi.Protocol.Linearization.LookupEvals (FVar F))
+    (feat : FeatureFlag → Bool) (inp : Inputs F) (α : F)
+    (htab : ∀ n, (inp.alphaPows n).val V = α ^ n) :
+    CircuitCompatible V (c := c) (inp.toEnv endo mds lk feat)
+      ((inp.evals.map (·.val V)).toEnv endo mds α (inp.beta.val V) (inp.gamma.val V)
+        (inp.jointCombiner.val V) (inp.vanishes.val V) (fun _ _ => 0)
```
martyall: How is it that ou can ahrd code the unormalized lagrange basis to all 0 here?

formal/pickles/Pickles/Linearization/Map.lean
```diff
@@ -0,0 +1,287 @@
+import Pickles.Linearization.Interpreter
```
martyall: check this module for unused bindings and maximally private bindings

formal/pickles/Pickles/Linearization/Map.lean
```diff
@@ -0,0 +1,287 @@
...
+decidable equality over polynomials.
+-/
+
+namespace Pickles.Linearization
+
+variable {R S : Type} (φ : R → S)
+
+/-- Push a carrier map through the machine state: both arrays cell by cell, position
+unchanged. -/
+def EvalState.map (s : EvalState R) : EvalState S :=
```
martyall: this should be defined alongside EvalState

formal/pickles/Pickles/Linearization/Map.lean
```diff
@@ -0,0 +1,287 @@
...
+  ⟨s.stack.map φ, s.store.map φ, s.position⟩
+
+/-- `φ` intertwines two environments: every operation, and every constant, agrees across
+it. The branch clause is conditional rather than pointwise — `ifFeature` receives thunks,
+so what is required is that related branches give related results. -/
+structure Compatible (e₁ : Env Id R) (e₂ : Env Id S) : Prop where
```
martyall: I question the need for a separate definition here, if you just assume \phi is an f-algrba homomoirphism you get all these for free

formal/pickles/Pickles/Linearization/Circuit.lean
```diff
@@ -0,0 +1,399 @@
...
+
+open Std.Do Snarky
+open scoped Kimchi
+
+variable {F c : Type} [Field F] [DecidableEq F] [BasicSystem F c] [ConstraintHolds F c]
+
+/-- A circuit environment computes a pure one, under a valuation. The affine fields agree
+on the nose; the constraint-emitting fields agree in the sense that any satisfying
+valuation reads their result correctly. -/
+structure CircuitCompatible (V : Valuation F)
```
martyall, line 54: Worth thinking about if you can define a ring strucutre on CircuitM F (FVar F) that makes evaluation an f-algebra hom that would obviate this definition

formal/pickles/Pickles/Linearization/Spec.lean
```diff
@@ -0,0 +1,182 @@
...
+token stream reads — `(endo, mds, α, e)` — so the interpreter environment that adjudicates
+the stream against it should be BUILT from those, not assembled independently. `toEnv` is
+that adapter, and it is the single definition the fixture driver and any reflection proof
+both go through.
+
+## What `toEnv` decides
+
+Four choices, none of them forced by the type:
+
+* **The feature predicate and the lookup evaluations are PARAMETERS**, following the
```
martyall: In general I could go through all the doc comments you've made in this PR and ask if any of these are really necessary. There are already style guides for lean comments which I can tell that you've broken several times

formal/pickles/Pickles/Linearization/Spec.lean
```diff
@@ -0,0 +1,182 @@
...
+/-! ## Naturality
+
+`toEnv` commutes with an `F`-algebra homomorphism, which is what lets the interpreter be
+run once over a polynomial algebra and read back at the field. The gate parameters are
+untouched — `φ` meets them only through `commutes`, since they enter via `algebraMap`. -/
+
+open Pickles.Linearization in
+/-- Mapping the evaluations and building the environment is building the environment and
+transporting it along `φ`. -/
+theorem toEnv_compatible {S : Type} [CommRing S] [Algebra F S] (φ : R →ₐ[F] S)
```
martyall: This has only one call cite, and again I would say that the F-algebra homomorphism would allow you to maybe skip the compatbile structure and just inline this

formal/pickles/Pickles/Reflect/Certificate.lean
```diff
@@ -0,0 +1,281 @@
...
+private def symValueP : MPoly Fp := evaluate (symEnv Pasta.pallasEndo symMds) fpTokens
+
+/-- **The certificate, `Fp` side.** -/
+private theorem fp_reflects :
+    symValueP = gateLinearization Pasta.pallasEndo symMds (xv 51) symEvals := by
+  native_decide
+
+/-- **The deployed `Fp` stream computes the gate linearization.** -/
+theorem evaluate_fpTokens (α β γ jc van : Fp) (e : Evals Fp) :
+    (evaluate (e.toEnv Pasta.pallasEndo symMds α β γ jc van (fun _ _ => 0) LookupEvals.zero
```
martyall: again I'm surprised that you can replace the lagrange by 0

formal/pickles/Pickles/Reflect/Poly.lean
```diff
@@ -0,0 +1,72 @@
...
+CompPoly supplies evaluation as a bare function (`aeval`) and as a ring homomorphism
+(`eval₂Hom`); this packages the two together with `aeval_C`, which is exactly the
+`commutes'` obligation.
+
+Unlike the instance above this may be noncomputable: it appears only in proofs, never in
+the certificate's computation. -/
+
+/-- Evaluation at a point, as an `R`-algebra homomorphism out of the polynomial algebra.
+`aevalAlgHom_X` and `aevalAlgHom_C` are its computation rules. -/
+noncomputable def aevalAlgHom {n : ℕ} {σ : Type*} [CommRing σ] [Algebra R σ]
```
martyall, line 45: THis is just that evaluation is an algebra hom, how is this not part of the standard library?

formal/pickles/Pickles/Linearization/Interpreter.lean
```diff
@@ -272,4 +272,100 @@ def evaluate [Monad m] (env : Env m F) (toks : Array PolishToken) : m F := do
   let s ← evalLoop env toks toks.size toks.size EvalState.init
   pure (topOrZero env s)
 
+
```
martyall, line 270: this module needs another maximally private check

formal/pickles/Pickles/Linearization/Spec.lean
```diff
@@ -136,47 +136,57 @@ def Evals.toEnv (endo : F) (mds : Kimchi.Gate.Poseidon.Mds F)
...
-  jointCombiner := rfl
-  beta := rfl
-  gamma := rfl
-  ifFeature f t₁ n₁ t₂ n₂ ht hn := by
-    simp only [Evals.toEnv]
-    split <;> assumption
+variable (endo : F) (mds : Kimchi.Gate.Poseidon.Mds F) (α β γ jc van : R)
+  (ulb : Bool → Int → R) (lk : LookupEvals R) (feat : FeatureFlag → Bool) (e : Evals R)
+
+@[simp] theorem Evals.toEnv_add (a b : R) :
```
martyall, line 138: check this module for maxmimum privacy

formal/pickles/Pickles/Linearization/Circuit.lean
```diff
@@ -347,26 +346,33 @@ def Inputs.toEnv [Field F] [DecidableEq F] [BasicSystem F c] (endo : F)
...
 lets a finite precomputed table (`precomputeAlphaPowers`) discharge it. The `α` here only
-seeds the base environment and is overwritten by the table. -/
+seeds the base environment and is overwritten by the table. The Lagrange basis is the
+constant zero on both sides; `evaluate_withUlb` lifts the circuit theorem to any gadget. -/
 theorem inputs_circuitCompatible [LawfulBasicSystem F c] {V : Valuation F} (endo : F)
     (mds : Kimchi.Gate.Poseidon.Mds F)
     (lk : Kimchi.Protocol.Linearization.LookupEvals (FVar F))
     (feat : FeatureFlag → Bool) (inp : Inputs F) (α : F) :
-    CircuitCompatible V (c := c) (inp.toEnv endo mds lk feat)
+    CircuitCompatible V (c := c) (inp.toEnv endo mds lk feat (fun _ _ => pure (.const 0)))
```
martyall: why are you already specializing here? feature flags aren't even specialized yet

formal/pickles/Pickles/Linearization/Interpreter.lean
```diff
@@ -372,4 +396,171 @@ theorem evaluate_withAlphaPow [Monad m] (env : Env m F) (g : Nat → F)
   simp only [evaluate, evalLoop_withAlphaPow env g toks hg]
   rfl
 
+
+/-! ## Where a disabled-features run goes
+
+With every feature disabled, `SkipIfNot` runs its second branch only, so the positions a
```
martyall: check this module for maximum privacy

formal/pickles/Pickles/Linearization/Interpreter.lean
```diff
@@ -372,4 +396,171 @@ theorem evaluate_withAlphaPow [Monad m] (env : Env m F) (g : Nat → F)
...
+  | some t => !t.isUlb
+  | none => true
+
+/-- `env` with its Lagrange-basis implementation replaced. -/
+def Env.withUlb (env : Env m F) (g : Bool → Int → m F) : Env m F :=
+  { env with unnormalizedLagrangeBasis := g }
+
+/-- A run with every feature disabled that visits no `unnormalizedLagrangeBasis` does not
+depend on the environment's implementation of it. -/
+theorem evalLoop_withUlb [Monad m] (env : Env m F) (g : Bool → Int → m F)
```
martyall: why would we now have 2 evalLoop theorems?

formal/pickles/Pickles/Linearization/Interpreter.lean
```diff
@@ -372,4 +396,171 @@ theorem evaluate_withAlphaPow [Monad m] (env : Env m F) (g : Nat → F)
...
+                  position := s.position + 1 + n + 1 + falseCount toks (s.position + 1 + n) }) :=
+            fun res => next _ _ rfl
+              (fun i hi => hvis i (List.mem_cons_of_mem _ (List.mem_append_right _ hi)))
+          rw [hfalse]
+          exact congrArg (bind _) (funext hcont)
+        | _ => exact next _ _ rfl (fun i hi => hvis i (List.mem_cons_of_mem _ hi))
+
+/-- The entry point, for a stream no disabled-features run reads
+`unnormalizedLagrangeBasis` in. -/
+theorem evaluate_withUlb [Monad m] (env : Env m F) (g : Bool → Int → m F)
```
martyall: why do we have a separate evaluate theorem just for this ?

formal/pickles/Pickles/Linearization/Spec.lean
```diff
@@ -187,6 +187,11 @@ variable (endo : F) (mds : Kimchi.Gate.Poseidon.Mds F) (α β γ jc van : R)
     (e.toEnv endo mds α β γ jc van ulb lk feat).ifFeature f t n
       = if feat f then t () else n () := rfl
 
+/-- Replacing the Lagrange basis of a `toEnv` is building it with the other one. -/
+theorem Evals.toEnv_withUlb (ulb₀ : Bool → Int → R) :
```
martyall: why is there a ulb variant ? why not make one genreal one

martyall (commented):
Something feels wrong about the way we treat the unnormalized lagrange basis vs the feature flags. The goal is that the library is maximally generic in that it only makes an assumption (eg that they all evaluate to 0) at the last possible moment (highest possible caller), similar to how we don't hardcode `fun _ => false` for all the feature flag gates

The fact that there are ulb and normal variants of many things is a huge red flag

PR #332, opened 2026-09-05
pickles: check_bulletproof reaches the wire's schnorrAt at both deployed curves

martyall:
## Summary

Finishes the algebra half of the `check_bulletproof` seam at the deployed fields: the circuit's success bit is the wire verifier's Schnorr check.

- **`checkBulletproof_wrap_spec` / `checkBulletproof_step_spec`** (`pickles/Pickles/CheckBulletproof.lean`): under any valuation satisfying the emitted constraints, `success` reads `1` iff `Bulletproof.Ipa.schnorrAt` holds at `IpaVesta.curve` / `IpaPallas.curve` — the `U` base the deployed group map's point up to the ordinate's sign, the challenges the endo-expansions of the 128-bit readings, the combined commitment `combineCommitments` over the kept bases at the expanded `ξ`, and `cip, b, z₁, z₂` the ladder witnesses' decodes cast into the scalar field. With `verifyWith_eq`, `success ∧ sgOk` is `verifyWith`.
- **Inputs that were open**: `groupMapCircuit_toGroup_spec` (group-map soundness up to sign under a genuine non-residue and no zero ordinate square) and its readings at both curves; `wrap_scale_reads` / `step_scale_reads` through `scaleFast{1,2}_spec` with `HasCurve.pallas` and the `{vesta,pallas}_ladderRegime` band lemmas (`snarky/.../Point.lean`); the `SchnorrPoint ↔ schnorrAt` bridge through `SWPoint.equivPoint` on both curves (integer action as the representative, Horner ↔ `combineCommitments`).
- The generic specs now also return the challenge count (needed to build the wire's `Vector` of challenges).
- `pasta`: `pallasAffineModule`, the twin of `vestaAffineModule`.
- `check_cs` uses the library group-map parameters (`groupMapParams{Vesta,Pallas}`) instead of its own copies.

Helpers and bridge lemmas are private; the public API is the two deployed specs and the two bridge iffs, registered in the roots and (as certificate-bearing, on the Pasta order/primality certificates) in the pickles axiom gate.

## Test plan

- [x] `make lean-build`
- [x] `scripts/check-style.sh`
- [x] pickles, pasta, snarky, bulletproof-pcs axiom gates
- [x] deadcode, shake, `runLinter` on Pickles and Pasta
- [x] `scripts/check_cs.lean` corpus (both `check_bulletproof_*` targets ✓)

## Update: `scale_fast` pins the top bit (mina + PureScript + Lean)

Proving the wrap side exposed a real slack in the production gadget: a 255-bit witness of a `Type1` shifted value `t` could be `t` or `t + |Fq|` (both fit for almost every `t`), and the second decodes to the scalar plus `2·(|Fq| − |Fp|)`, so `check_bulletproof` and `ft_comm` on the wrap side scaled by a prover-chosen offset. The remedy is the one `scale_fast2` already applies above its split: pin the top bit to zero when `num_bits` reaches the field width.

- **mina** (`cip-wrap-dump` @ `c0dc591e54`, submodule bumped): `scale_fast` asserts the top bit is zero at full width. An `Equal` against the cached zero, so wiring only, no gate rows; kept on one line so `__LOC__` labels are undisturbed.
- **PureScript** `scaleFast1` and **Lean** `scaleFast1` mirror it; `scaleFast1_spec` / `_complete` carry the halved bound.
- **Lean downstream**: the wrap ladder witness is canonical, `wrapAlias` and the `ε` terms are gone, and `checkBulletproof_wrap_spec` states `schnorrAt` at `wrapDecode V x` outright, symmetric with the step side.
- **Fixtures**: regenerated from the patched submodule (`dump_circuit`, the top-level drivers, `dump_chunks2`). Twelve dumps changed content: `check_bulletproof_wrap`, `ftcomm_wrap`, `ivp_wrap`, `wrap_verify{,_n2}`, the six `wrap_main_*`, `chunks2_wrap_main`, `var_base_mul_step`; plus the side-loaded NRR proof/VK (the wrap VK changed). A separate commit refreshes stale `_cached_constants`/`_labels` companions that had drifted from the submodule independently of this change.
- **Checks**: circuit-diffs 110/110, NRR 4/4, `check_cs` corpus clean, full Lean gates green.

Excluded scalars: shifted values in `[2^254, |Fp|)`, a `2^-129` fraction of the field, which honest provers will not hit. Upstream PR to follow once settled.


martyall:
Follow-up on the ladder-witness question (commit above): the deployed conclusions no longer quantify over witnesses.

- **Step**: the `Type2` half is unpacked in 254 bits and `2^254 < |Fp|`, so the witness is canonical; `checkBulletproof_step_spec` states `schnorrAt` at `stepDecode V x` outright.
- **Wrap**: a 255-bit witness of a `Type1` shifted value `t` is `t` or `t + |Fq|` (both fit for almost every `t`), and the second decodes to the scalar plus `wrapAlias = 2·(|Fq| − |Fp|)` in `Fp`. `checkBulletproof_wrap_spec` now states `schnorrAt` at `wrapDecode V x + ε` with a prover-chosen `ε ∈ {0, wrapAlias}` per scaled scalar.

The wrap slack is the production gadget's, not the model's: `scale_fast_unpack` (`plonk_curve_ops.ml`) witnesses the bits with `exists` and asserts only `n_acc = scalar` in the circuit field, and `Impls.Wrap.Other_field.check` excludes the degenerate representatives modulo `|Fp|` (`forbidden_shifted_values`) but not the circuit-field alias. So the wrap circuit implements the Schnorr check at `cip + ε_cip`, `b + ε_b` for a fixed, prover-known offset. Whether that matters downstream is a Fiat–Shamir question (the offset is fixed before the challenges that bind the evaluations), flagged here rather than settled.

PR #337, opened 2026-09-09
pickles: the verifier recomputes the message digests and rebuilds the accumulator list

martyall:
Two commits on the out-of-circuit verifier (`Pickles.Verify`), file-disjoint from #332 and #334; the tests below ran on this branch against main and, earlier, on the same changes on top of #334's tree.

## Summary

- **The verifier recomputes both message digests from the wrap key.** `VerifiableProof` no longer carries the two digests the prover supplied; it carries what they hash — the application state, the previous proofs' challenge-polynomial commitments and their expanded wrap challenges — and `messageDigests` recomputes both from the verifier's own wrap key, as OCaml `verify.ml` does (`hash_messages_for_next_step_proof ~dlog_plonk_index:key.commitments ~app_state`; `Wrap_hack.hash_messages_for_next_wrap_proof` over the dummy-padded challenges). Both land in the wrap public input, so a proof whose step circuit hashed any other key or state fails the kimchi check. Before, a prover-chosen digest bound the chain to a key and state of the prover's choosing. `extractWrapVKForStepHash` moves to `VerificationKey`; codecs and the fixture loader carry the new fields.
- **The verifier rebuilds the wrap proof's accumulator list from the messages.** The terminator verified the wrap proof with whatever `prev_challenges` the prover had stored in the proof object (`verifyOpeningProofsBatch` passed an empty override, which keeps the stored list), so a prover could open the wrap proof at a self-consistent list unrelated to the `sg` it hashed. `wrapAccumulators` now builds the list OCaml's `verify.ml` builds — step-message commitments and wrap-message expanded challenges, front-padded to `PaddedLength` with the dummy wrap accumulator — and `verifyOpeningProofsBatch` injects it per entry. `Verifier` carries the dummy wrap `sg`, so `mkVerifier` takes the Pallas SRS.

Tests: a proof presented for another application state, another previous `sg`, or other previous wrap challenges is rejected (`NoRecursionReturn`, `SimpleChain`, `VerifyNrrSpec`); `proofPrevChallenges` reads a proof's stored list back, and `VerifyFixturesSpec` asserts the rebuilt list equals what OCaml's prover stored on all six side-loaded fixtures (mpv = 1 and 2, base cases included).

## Test plan

- [x] `purs compile` type-check of pickles with its tests, `purs-tidy check` on the touched files
- [x] `spago test -p pickles` — 21/21, on this branch against main (`ffe562a6`) and on the same changes over #334's tree


packages/pickles/src/Pickles/Verify.purs
```diff
@@ -345,15 +370,57 @@ toVerifiable (CompiledProof p) =
...
+messageDigests verifier vp =
+  let
+    stepProofs = Array.zipWith
+      (\sg expandedBpChallenges -> { sg, expandedBpChallenges })
+      vp.prevChallengePolynomialCommitments
+      vp.oldBulletproofChallenges
+
+    step = Vector.reifyVector stepProofs \proofs ->
+      hashMessagesForNextStepProofPure
+        { stepVk: extractWrapVKForStepHash @1 verifier.wrapVK
```
martyall: what's up with magic number 1, because wrap always has 1 chunk?

martyall: Always 1: a wrap circuit's domain is 2^13/2^14/2^15 by `max_proofs_verified` (`common.ml` `wrap_domains`) and never exceeds the Tock SRS, 2^15, so every wrap-VK commitment is one chunk. OCaml pins the same at `num_chunks_by_default = 1` (`step_main.ml:347`) and `verify.ml` hashes each commitment as `[| x |]`. Now named: `type WrapVkChunks = 1` in `Pickles.Types`, used here and at the three step-prover sites that had `@1`.

PR #369, opened 2026-09-16
pickles: rules return one statement type per slot (#363)

### LINKED ISSUE #363: pickles: rules cannot have per-slot statement types — RuleOutput collapses the indexed product to a power
A rule's slots are supposed to be able to carry different tags, and therefore
different statement types. The type-level spec language in this repo says so,
OCaml says so, and the spec says so — but `RuleOutput` forces every slot in a
rule to share one statement type.

## What the definition says

From `kimchi-spec.tex`, Definition (Rules): a rule at a compiled tag `T` with
`n` slots, slot `i < n` carrying tag `T_i`, is a circuit whose public output is

```
Output_T × ∏_{i<n} (S_{T_i} × Bool)
```

That is an *indexed* product — `S_{T_i}` varies with `i`. The family figure is
explicit about it: the rule at `A` has an External slot to `B`, a Self slot,
and a side-loaded slot carrying `σ`, three different statement types in one
rule ("`C` has `σ`'s statement type").

## What the type-level spec language says

`Pickles.Step.Slots` builds a genuinely heterogeneous carrier from the spec:

```purescript
instance SlotStatementsCarrier rest restValCarrier =>
  SlotStatementsCarrier (Slot n statement /\ rest) (statement /\ restValCarrier)
```

one `statement` per slot, threaded into a nested tuple. The module docstring
even gives a mixed example.

## Where it collapses

`Pickles.Step.Main`:

```purescript
type RuleOutput n prevInput output =
  { prevPublicInputs :: Vector n prevInput
  , proofMustVerify :: Vector n (BoolVar StepField)
  , publicOutput :: output
  }
```

`Vector n prevInput` — one `prevInput` for all `n` slots.

The `Vector` is not an incidental container choice. `Step/Main.purs:845-847`
encodes every slot's statement through a single `CircuitType` instance:

```purescript
prevInputVar = prevPublicInputs !! i
input = buildVerifyOneInput pw (varToFields @StepField @prevInputVal prevInputVar) …
```

and `StepRule`/`StepRuleAt` carry exactly one such instance for the whole rule
(`CircuitType StepField prevInputVal prevInput`). So the homogeneity is forced
by there being one encoding in scope per rule, not by the container type.

## Net effect: the two axes have diverged

| | shape | carries |
| --- | --- | --- |
| `AsProver … valCarrier` | heterogeneous (`s₁ /\ s₂ /\ Unit`) | what the prover knows about each slot |
| `RuleOutput.prevPublicInputs` | homogeneous (`Vector n prevInput`) | what the rule returns as wires, and what gets hashed |

Heterogeneity survives on the advice path and dies on the verified path — the
one that matters. The `∏` is implemented as a power.

## This is a PureScript narrowing, not a protocol restriction

OCaml returns a heterogeneous list indexed by the per-slot var types
(`inductive_rule.ml`):

```ocaml
type ('prev_vars, 'widths, 'public_output, 'auxiliary_output) main_return =
  { previous_proof_statements :
      ('prev_vars, 'widths) H2.T(Previous_proof_statement).t
  ; public_output : 'public_output
  ; auxiliary_output : 'auxiliary_output
  }
```

So the spec, OCaml, and this repo's own type-level slot spec all say
heterogeneous; only the PureScript value-level return says otherwise.

## Impact

Currently unexercised. Every multi-slot rule in the tree is homogeneous
(`Slot 2 TxnStmt /\ Slot 2 TxnStmt /\ Unit` in the transaction example), and
every side-loaded fixture has at most one slot, so nothing today needs a rule
that mixes a Self slot with a side-loaded one — which is precisely the shape
the spec's own worked example has.

It does mean a rule cannot be written for that shape, and that a reader of
`Pickles.Step.Slots` will reasonably believe it can be.

## Shape of a fix

Make `prevPublicInputs` a `SlotStatementsCarrier`-derived carrier with a
per-slot `CircuitType`, rather than a `Vector` under one instance. That is a
change to `stepMain`'s slot loop and to the `StepRule` constraint set, not a
type-alias tweak.

## Incidental

`Pickles.Step.Slots`'s module docstring writes `Slot 1 1 (StatementIO Stmt)`
with three arguments, but `Slot :: Int -> Type -> Type` takes two — stale from
the `nc` removal.

martyall:
Closes #363.

## What changes

A rule's slots may carry different tags, so different statement types. `RuleOutput` returned `Vector n prevInput` under one `CircuitType`, and `compileMulti` threaded a single `prevInputVal` through every branch, so neither a rule nor an application could mix statement types.

- **Rules:** each slot returns `PrevStatement { publicInput, proofMustVerify }` at the variable type of the statement its `Slot n stmt` declares. `RuleOutput prevsSpec output = { prevs :: Prevs prevsSpec, publicOutput }`.
- **One place for slot types:** `StepRule prevsSpec inputVal input outputVal output`. The body opens advice with `prevValues` and packs its return with `toPrevs`; both compute their tuple from the spec by fundep.
- **Compile machinery:** `prevInputVal` removed from `CompilableRulesSpec`, `CompilableRulesSpecShape`, `mkRuleEntry`, `compileMulti` and the step functions. `RulesCons mpv prevsSpec rest` no longer restates the value carrier.

Level 3 of the issue (`wrapVkChunks` in `BuildSlotVkSources`) is out of scope: wrap chunk counts are fixed at 1 in OCaml too, so no slot can differ.

## Bug fixed on the way

`finalize_other_proof` in the step circuit did not deduplicate a slot's candidate step domains. An `External` slot inside a multi-branch compile, or two branches with equal step domain sizes, set two which-bits and failed `prevs_verified`. Candidates are now sorted by `log2` and deduplicated, as the step verifier's `domain_for_compiled` does.

## Test

`HeterogeneousPrevs`: a child with a 1-field input statement, and a two-branch application with a 2-field output statement (a base branch with no slots; a branch with an `External` slot at the child's statement and a `Self` slot at its own). The chain must produce `(0,0)`, `(1,7)`, `(2,14)` and verify.

## Checked locally

- `circuit_diff` matches OCaml on all 10 step fixtures and `finalize_other_proof_step_circuit`.
- Narrowed prove tests: `HeterogeneousPrevs`, `TreeProofReturn`, `TwoPhaseChain`, `PaddedWideSlots`.
- Not run locally: the full `ps:pickles` and `ps:example` suites. `pickles-bench` is its own workspace and was edited without a type-check.


packages/pickles/src/Pickles/Step/Main.purs
```diff
@@ -650,16 +644,17 @@ stepMain
...
-  { prevPublicInputs, proofMustVerify, publicOutput, perSlotVkSources } <-
+  { prevs, publicOutput, perSlotVkSources } <-
     label "rule_main" do
       perSlotVkSources <- buildSlotVkSources @cell @prevsSpec @WrapVkChunks perSlotVkBlueprints sideloadedVkCarrier
       -- The rule reads previous proofs' statements through this
       -- deferred getter, forced only inside the rule's own `exists`
       -- bodies.
-      result <- rule (pure advice <#> \(StepAdvice r) -> r.prevAppStates) publicInput
+      result <- rule
+        (pure advice <#> \(StepAdvice r) -> mkPrevValues @prevsSpec r.prevAppStates)
```
martyall, line 654: this is sort of a small to me that we need an advice monad only to return a pure value

Issue #333, opened 2026-09-08
pickles: side-loaded verification key is an unbound witness; rules cannot constrain it


martyall:
## Summary

A side-loaded prev slot's verification key is an unconstrained witness of the step circuit. Pickles verifies the prev's wrap proof against whatever key the prover supplies, and nothing binds that key to the statement. A rule cannot bind it either, because the rule never receives the key variable.

## Where

- `packages/pickles/src/Pickles/Step/Main.purs`, `buildSlotVkSources`, `SideLoaded` instance: the key is allocated with `exists (pure (projectVk headCell))` and wrapped as `SideloadedExistsVk`.
- The rule is called as `rule (pure advice <#> \(StepAdvice r) -> r.prevAppStates) publicInput`; the per-slot VK sources are not passed to it.
- `hashMessagesForNextStepProofPure` and the in-circuit digest absorb only the shared Self wrap key (`exists_wrap_index`), never a side-loaded key.
- `test/Test/Pickles/Prove/SideLoadedMain.purs` supplies `childVK` via `sideloadedVKs` and asserts nothing about it beyond the proof verifying.

## Consequence

Knowledge soundness still holds formally: an accepted proof yields a derivation whose side-loaded subtree is at *some* application, namely the one owning the supplied key. If the prover compiled that application itself, the slot certifies nothing about the intended child. Concretely, `SideLoadedMain`'s parent proof should verify with any child key, not only the real one.

## How Mina does it

Binding is the rule's obligation. In `mina/src/lib/transaction_snark/transaction_snark.ml` (around line 1448) the zkApp path computes `Zkapp_account.Checked.digest_vk vk`, asserts it equal to the account's stored `verification_key` hash, and only then calls `Pickles.Side_loaded.in_circuit (side_loaded 0) vk`. The key variable is the rule's to `exists` and constrain.

## Proposed fix

Option 1 (matches Mina; keeps circuit-diff fixtures): hand the allocated side-loaded key variable to the rule, so the rule can hash it and constrain the hash against its statement. E.g. extend the getter the rule receives (`prevAppStates` / the `valCarrier`) with the per-slot side-loaded VK var, or pass `perSlotVkSources` into the rule.

Option 2 (framework binding; diverges from Mina's statement layout and fixtures): absorb a hash of each side-loaded key into `messagesForNextStepProof`.

Either way, add a negative test: `SideLoadedMain` proving with a wrong child key must fail once the rule binds it.

## Paper

[...]

PR #374, opened 2026-09-17
pickles: a slot's kind decides where its verification key comes from (#333, phase 1)

martyall:
Phase 1 of #333: types and plumbing. `bindVk` and the negative test follow in later commits on this branch.

## Problem

A side-loaded slot's verification key was an unconstrained witness that the framework allocated (`buildSlotVkSources`), and the rule never received it. Pickles verified the previous proof against whatever key the prover supplied, and no rule could tie that key to its own statement. In Mina the binding is the rule's obligation (`transaction_snark.ml`: the rule creates `vk`, asserts `digest_vk vk` equals the account's stored hash, then calls `Side_loaded.in_circuit`).

## Change

- **Slots carry a kind.** `SlotOf (k :: SlotKind) n stmt`, with `Compiled` and `SideLoaded`, plus the synonyms `type Slot n stmt = SlotOf Compiled n stmt` and `type SideLoadedSlot n stmt = SlotOf SideLoaded n stmt`. Existing specs written as `Slot n stmt` are unchanged.
- **A side-loaded slot's advice** is `{ statement, verificationKey }` rather than the bare statement; a compiled slot is untouched.
- **A side-loaded slot's return** is `SideLoadedPrevStatement { publicInput, proofMustVerify, verificationKey :: BoundVk }`. The step circuit verifies that slot against the returned key, and `buildSlotVkSources` no longer allocates one.
- **`BoundVk` is opaque.** Its constructor lives in `Pickles.Sideload.BoundVk.Internal` as `unsafeUnboundVk`, used only by the circuit-diff harness and the test that reproduce an OCaml rule which binds nothing. Once `bindVk` lands, a rule that has not bound its key cannot construct one.
- **`mkRuleEntry` refuses** a rule whose declared slot kinds disagree with the `SlotWrapKey` values supplied for those slots.

## Not yet in this branch

- `bindVk` and the `digestVk` port of Mina's `Zkapp_account.Checked.digest_vk` (`Random_oracle` with the `MinaSideLoadedVk` prefix over the key's `to_input`).
- A test that proves with the wrong child key and fails.

## Checked locally

- `pickles` (source and tests), `pickles-circuit-diffs` and `example` type-check clean; lint passes.

## Circuit parity: CI's job here

Local `circuit_diff` runs could not certify parity: this checkout's `mina` submodule is at `a75fc7da17` ("assert_split_below spends one range check per split") rather than the `f2265df310` this branch pins, and the OCaml exports under `circuits/ocaml/` were regenerated from it. Four step fixtures differ by exactly the extra range check per scalar-challenge split, per slot — the OCaml side moving, not this change. CI checks out the pinned submodule, so its circuit-diff run is the real gate.


packages/pickles-circuit-diffs/src/Pickles/CircuitDiffs/PureScript/StepMainSideLoadedMain.purs
```diff
@@ -72,7 +72,12 @@ sideLoadedMainRule
   -> Snarky StepField (KimchiConstraint StepField) r
        (RuleOutput SideLoadedMainPrevsSpec Unit)
 sideLoadedMainRule getPrevStates appState = do
-  prev <- exists $ getPrevStates <#> prevValues <#> \(StatementIO p1 /\ _) -> p1.input
+  prev <- exists $ getPrevStates <#> prevValues <#> \(slot /\ _) ->
+    let StatementIO p1 = slot.statement in p1.input
+  -- The OCaml rule allocates the key here, between the prev statement
+  -- and the base-case comparison, and hands it to
+  -- `Side_loaded.in_circuit` without binding it to anything.
+  vk <- exists $ getPrevStates <#> prevValues <#> \(slot /\ _) -> slot.verificationKey
```
martyall: why witness it again,  just take everything you need from the first witness.

packages/pickles/src/Pickles/Prove/Compile.purs
```diff
@@ -1526,9 +1536,24 @@ class SlotWidths (prevsSpec :: Type) where
...
 
-instance (Reflectable n Int, SlotWidths rest) => SlotWidths (Slot n stmt /\ rest) where
+instance (Reflectable n Int, SlotWidths rest) => SlotWidths (SlotOf k n stmt /\ rest) where
   slotWidthsOf _ = Array.cons (reflectType (Proxy @n)) (slotWidthsOf (Proxy @rest))
 
+-- | Whether each slot is side-loaded, in slot order, read back from
+-- | the spec's `SlotOf` kinds. `mkRuleEntry` checks the answer against
+-- | the `SlotWrapKey` the caller supplies for that slot.
+class SlotKinds (prevsSpec :: Type) where
+  slotKindsOf :: forall proxy. proxy prevsSpec -> Array Boolean
```
martyall: the use of booleans here is a hack, why not a custom enum

Issue #373, opened 2026-09-17
formal: satisfiability of the two circuit halves on deployed proofs


martyall:
## Summary

The two-halves capstones (`twoHalves_kimchiVerify_vesta`, `twoHalves_kimchiVerify_pallas`) relate the circuits' success bits to `kimchiVerify` for any valuation that satisfies the circuits and whose reads hold. They say nothing about whether a real proof produces such a valuation. There are no completeness theorems at this level, and getting them is not a realistic target. What we can have instead is a satisfiability test on deployed data: run the modeled circuit halves through the Lean prover interpreter on the inputs of a real proof and check that the resulting valuation satisfies every constraint and that the success bits read 1, while `kimchiVerify` on the same proof also accepts.

This is the concrete complement to `solve_complete`, and the same kind of gate as `kimchi/scripts/check_kimchi_verifier.sh`, which runs the Lean verifier on deployed proofs. It tests something neither the theorems nor `check-cs` can: that the Lean model's advice, the witness code, agrees with the PureScript prover on deployed data. `check-cs` compares constraint systems only.

## What exists

- `formal/kimchi/fixtures/kimchi_proof_pallas_pickles.json`: a deployed wrap proof (`simple_chain`'s second wrap) that the Lean `kimchiVerify` accepts, produced by `tools/fixture-dump`'s `kimchi_proof_dump_pickles` from the side-loaded fixture plus `simple_chain/wrap1/lean_inputs.json`.
- `Test.Pickles.Sideload.LeanInputsSpec` writes `lean_inputs.json`: the wrap public input (expanded deferred values, the two recomputed digests) and the accumulator list the proof is opened at. That is the verifier's view of the proof, not yet the circuit's.
- The Lean circuit halves `incrementallyVerifyProof` (group side) and `finalizeOtherProofStep` / `finalizeOtherProofWrap` (scalar side), with the `Snarky` prover interpreter and `solve` over them.
- `packages/pickles/test/fixtures/proof-cache/*.json`: every test rule's proofs keyed by (vk, public input), one entry per distinct circuit shape after a clean regeneration.

## What to build

1. **A circuit-input dump.** Extend the `LeanInputsSpec` mechanism (or a sibling spec) to write, per proof, what the two halves take as witness: the unfinalized claims, the evaluations, the previous challenges, the fq-sponge digest, the mask where the step side varies it, and the group-half's public input commitment data. Decimal strings, as the Lean fixtures already use.
[...]

PR #375, opened 2026-09-17
snarky: the prover's witness table is a dense array

martyall:
## Why

`Assignments F` was `Variable → Option F`. A write was a closure over the previous table, so a read walked one closure per assigned variable and a prover run over `n` variables cost **O(n²)**.

PureScript settled this already. `packages/snarky/src/Snarky/Backend/Assignments.purs` is a dense `STArray` indexed by the variable, and its docstring records both the reasoning — *"variables allocate sequentially, so writes are appends. O(1) write-once inserts and O(1) witness lookups versus the old `Map`'s O(log n) path-copying on both"* — and a measurement for a later refinement. The Lean model was therefore slower than the representation PureScript **discarded**: O(n) against the old `Map`'s O(log n), against the current array's O(1). `formal/docs/snarky-ps-alignment.md` signs this tree off as aligned module by module, so it was a divergence, not just a slow choice.

`formal/CLAUDE.md` already names the hazard in general terms: *"state threaded through executable folds must be concrete data (tuples, structures) — the compiler eta-expands function-valued definitions, making folds exponential."* `prove` is exactly such a fold.

## What

`Assignments F` becomes a structure over `Array (Option F)`, read through a `CoeFun` to `Variable → Option F`. Writes pad to the index and set in place; on the sequential path — the only one `prove` takes — that is a single push.

## Measured

Build a table of `n` entries, read all of them back:

| n | function | n | array |
|---|---|---|---|
| 2,000 | 467 ms | 10,000 | 5 ms |
| 4,000 | 1,875 ms | 20,000 | 10 ms |
| 8,000 | 7,479 ms | 40,000 | 20 ms |

4× per doubling against 2× — quadratic against linear, **~1,900× at n = 8,000** and widening.

## Blast radius

Two files, +46/−7.

- `formal/snarky/Snarky/Assignments.lean` — the representation. Three private helpers carry it (`padTo`, `lt_size_padTo`, `getElem?_padTo`); every one of the 12 lemmas keeps its statement verbatim, because the API was already stated by applying the table (`a v = some x`, `(a v).isSome`).
- `formal/snarky/roots.txt` — one line. Elaboration unfolds the coercion straight into the private `lookup`, so no term mentions the `CoeFun` instance and the dead-code gate cannot see it. Same category as `Snarky.AsProver.instMonad` twelve lines above, and as `Kimchi.Index.instDecidableSatisfies`, which carries its reason inline the same way.

**Nothing downstream changed.** All 64 modules importing `Snarky.Assignments` across `Snarky`, `Pickles` and `Schnorr` build untouched — including `AsProver.run`'s `match env v`, the hot-path read.

## Checks

- `lint deep=true`: axiom gates for all five packages, env linters for all four libraries, shake, dead-code — green; commit gate open.
- `check_dependents Snarky.Assignments`: 64/64 modules, no errors.
- `ps:pickles-circuit-diffs`: full suite, 109 comparisons, no mismatch.

No constraint system can move: `build : CircuitM F c α → ℕ → Built c α` and `Built` carries `result`/`nextVar`/`constraints`, no assignment. The table exists only on the `prove`/`solve` path. The circuit-diffs run is confirmation, not a dependency.

## Not in this PR

- `formal/docs/snarky-ps-alignment.md:74` still describes this as *"present as `Nat → Option F`"*, which is now false. That file records a completed sign-off walk rather than live status, so I left the edit for a decision.
- A harness gap found on the way: `snarky-agent test ps:pickles-circuit-diffs` invokes spago directly and so skips the `fixtures-unpack` prerequisite that `make test-pickles-circuit-diffs` carries. After a branch switch the untracked `.json` working copies go stale — 177 of mine were — and the test then reports a detailed, entirely false mismatch.

## Context

This unblocks running `prove` on a real circuit, which is what l-adic/snarky#373 needs and which nothing in the tree has ever done.


martyall:
Follow-up: the same before/after on a real circuit, rather than the synthetic table in the description.

`Pickles.finalizeOtherProofStep` at the step side's deployed parameters — 1,407 constraints, **14,128 variables** — built and proved through `Snarky.prove`, compiled (`lake exe`), identical input and identical output bits in both runs. The only difference is this PR:

| `Assignments` | `prove` |
|---|---|
| `Variable → Option F` (main) | **21,715 ms** |
| `Array (Option F)` (this PR) | **600 ms** |

**36×**, and the gap widens with variable count because the old shape is quadratic in it.

Two consequences worth recording:

- Nothing in the tree had ever run `prove` on a verifier-sized circuit. At 21.7 s for the *smaller* of the two halves, that is not an accident of priorities — it was impractical.
- The group half is roughly ten times the variables. On a quadratic curve that is minutes per run, so l-adic/snarky#373's satisfiability check was not merely slow before this, it was out of reach.

Measured with a driver added on the follow-up branch, not part of this PR.

PR #376, opened 2026-09-17
pickles: the proof cache records a chain's links and verification keys

martyall:
The pickles test suite's proof cache becomes a record a downstream consumer can verify a chain from, not only a memo the prover reads back.

## What changes

- **Every entry carries its verification key**, in the natural PureScript encoding (`vkRawToJson`: camelCase, uncompressed `[x, y]` little-endian hex points). Downstream decoders deal with what they get.
- **The store is keyed by the key's digest**, `VerifierIndex::digest()` (the value the verifier absorbs first), computed in PureScript by `Pickles.VerificationKey.verifierIndexDigest` and checked against the recorded fixture value in `LeanInputsSpec`. `loadStore` drops buckets whose key is not a decimal digest, so a store written under the old keying cannot survive beside the new one.
- **Entries link the chain**: a wrap entry records the step proof it wrapped (`step: {vkDigest, publicInput}`); a step entry records, per slot, the wrap proof it verified there (`prevs`, `null` on a base-case slot). Both are data the prover has at the write site; `StepRef` becomes `ProofRef`.
- The cache under `packages/pickles/test/fixtures/proof-cache/` is regenerated from empty (58 entries, 11 files).
- The `snarky-kimchi` round-trip test keys by a decimal stand-in, as the store requires since the digest keying.

## Why

`formal/` runs the Lean circuit halves and `kimchiVerify` on these proofs (stacked PR). That needs, per proof, its key, its public input, and which proof it is built on — none of which the old `{vk-json -> public-input -> proof}` memo made recoverable.

## Testing

`ps:pickles` (243 s, regenerates the cache) and `ps:snarky-kimchi` green.


packages/snarky-kimchi/src/Snarky/Backend/Kimchi/ProofCache.purs
```diff
@@ -109,58 +132,62 @@ fieldStr = show <<< toBigInt
...
+  -> ProofRef
   -> Effect Unit
-setVestaProof cache vk pis proof =
-  setStr cache (vestaProofVkKey vk) (piKey pis) (vestaProofToSerdeJson proof)
+setVestaProof cache vkDigest vk pis proof step =
+  setEntry cache vkDigest (piKey pis)
+    { vk: vestaVerifierIndexJsonKey vk
+    , proof: vestaProofToSerdeJson proof
+    , step: Just step
+    , prevs: Nothing
```
martyall: why always Nothing here?

packages/pickles/src/Pickles/Prove/Compile.purs
```diff
@@ -502,6 +502,7 @@ consShapeCompileData cfg selfStepDomainLog2s headSlot restShape =
       , crs: cfg.srs.vestaSrs
       , debug: cfg.debug
       , proofCache: cfg.proofCache
+      , prevProofs: []
```
martyall: why hard coded []?

packages/pickles/src/Pickles/Prove/Compile.purs
```diff
@@ -1287,6 +1301,7 @@ instance CompilableSpec Unit Unit 0 Unit Unit Unit Unit where
         , crs: cfg.srs.vestaSrs
         , debug: cfg.debug
         , proofCache: cfg.proofCache
+        , prevProofs: []
```
martyall: why hard coded. [] ?

packages/pickles/src/Pickles/Prove/Compile.purs
```diff
@@ -1327,6 +1342,7 @@ instance CompilableSpec Unit Unit 0 Unit Unit Unit Unit where
               )
         , challengePolynomialCommitments: Vector.nil
         , baseCaseWrapPublicInputs: Vector.nil
+        , prevProofRefs: []
```
martyall, line 1344: why hard coded [] ?

PR #377, opened 2026-09-17
formal: the circuit halves are satisfied on the proofs the PureScript suite cached (#373)

### LINKED ISSUE #373: formal: satisfiability of the two circuit halves on deployed proofs

martyall:
Closes #373. Stacked on the proof-cache PR (#376).

The two-halves capstones relate a circuit's success bits to `kimchiVerify` for any table that satisfies the circuit; they say nothing about whether a real proof produces such a table. `lake exe check-halves` (from `formal/`) runs every modelled half through the Lean prover interpreter on the inputs of real proofs — the advice code, which nothing else in the tree exercised — and decides satisfiability on the table it produced:

| half | pairs | per run | what is checked |
| --- | --- | --- | --- |
| step `finalize_other_proof` | wrap → the step proof it wrapped | 1.5 s | prove completes, table satisfies (`Index.Satisfies`, decided), `finalized` and its four conjuncts read 1 |
| wrap `finalize_other_proof` | step slot → the wrap proof it verified | 1.1 s | same |
| step `verify` (group half of a wrap proof) | same pair | 6.4 s | prove completes, table satisfies, the opening's `success` reads 1; the digest and round-challenge assertions are the gadget's own constraints |
| wrap `incrementally_verify_proof` (group half of a step proof, conditional sponge) | wrap → step pair | 12.9 s | same, with the verify block's assertions |
| `kimchiVerify` | every proof | 10.5 s | accepts, against the project's SRS |

On `SimpleChain.json`: 23 runs over 10 proofs, all green, 200 s compiled for the whole corpus. Wired into CI as its own lane (`halves` in `lean.yml`): the build job compiles the driver alongside `check-cs`, the lane fetches the SRS and runs it. `HALVES=` and `LIMIT=` narrow a run.

## Pieces

- `KimchiFixture.Cache` — the PureScript proof cache decoded into the wire records: the key (camelCase JSON, uncompressed points), the proof (arkworks serde, compressed points via Tonelli–Shanks), the links. Validated field by field against the Rust-produced kimchi fixtures.
- `BulletproofFixture.SRSLoader` — `srs-cache/{vesta,pallas}.srs` read behind one function, `loadSRS` (MessagePack of arkworks-compressed points), cut to a proof's round count. Validated against the fixtures' `srs_g`/`srs_h` on both curves. Isolated so it can be swapped for another reader.
- `lagrangeBasisCached` — the Lagrange basis a wire key needs, computed in closed form (`omega^(-ik)/n` against the first `n` generators, one MSM each), validated against the Rust-computed `lagrange_basis`, memoised under `formal/lagrange-cache/` (gitignored).
- `PicklesFixture` — the harnesses `check_cs` and this driver share: the dump layouts (`Fop`, parametrised by round count — real wrap proofs have 15 rounds where the dump fixes 16), the named input bundles (`FopInput`, `FopWrapInput`, `GroupStepInput`, `GroupWrapInput`) whose encodings are the flat layouts, both sides' `x_hat` tables and key-as-constants helpers (`Group`), and `Satisfies` — the reduce → assemble → witness → `Kimchi.Fixture.PS.build` → `decide` path.
- `Snarky.Assignments` as a dense array landed separately (#375): `prove` on the step half went from 21.7 s to 0.6 s.

## Found along the way

- `Index.build?` decided `Function.Bijective` on the wiring with the generic `Fintype` instance — quadratic in the 7·n cells, 112 s of a 117 s group-half run at n = 8192. It now builds the wiring's predecessor table and decides the two round trips (linear), concluding bijectivity from `Function.bijective_iff_has_inverse`; same field, same law. `check_cs` and the index fixture check pay the same bill today.
- The dead-code gate imports fixture modules by name, so `KimchiFixture.Cache` was outside the audit; it and `PicklesFixture` are imported now, with their script-consumed surface rooted.
- The accumulator slots are padded in front: the mask reads slot `i` as "at least `2 - i` proofs"; on the wrap side the packed branch data holds `mask_i` at bit `1 - i` and `sg_old` is listed in reverse slot order. The first step-half run absorbed the wrong slot's challenges and read every sponge-dependent bit 0.

## Not in this PR

A negative control (a corrupted proof reading 0) and the other cache files (multi-chunk proofs need `nc > 1` harnesses).


formal/pickles/PicklesFixture/Fop.lean
```diff
@@ -0,0 +1,87 @@
+import PicklesFixture.Layout
```
martyall: I think this whole module is not needed

formal/pickles/PicklesFixture/FopInput.lean
```diff
@@ -0,0 +1,135 @@
...
+namespace PicklesFixture
+
+open Snarky Snarky.Kimchi CompElliptic.Fields.Pasta
+
+/-- The step side's `finalize_other_proof` input, by name. The field order is the dump's
+cell order: the claims at 0–25, the mask at 26–27, `domain_log2` at 28, the evaluation
+block from 29, `ft(ζω)`, the two previous-challenge vectors, and the digest last. -/
+structure FopInput (α : Type) where
+  /-- `α, β, γ, ζ` at 0–3, the five shifted claims at 4–8, `ξ` at 9, 16 challenges at 10–25. -/
+  claims : Vector α 26
```
martyall: Do we not have structured data to use laying around rather than just supplying vectors?

formal/pickles/PicklesFixture/FopInput.lean
```diff
@@ -0,0 +1,135 @@
...
+
+/-- The step harness on the named bundle at the dump's domain. -/
+def fopStepOn (input : FopInput (FVar Fp)) : CircuitM Fp C (Pickles.FopOutput Fp) :=
+  fopStepHarness (CircuitType.varToFields (val := FopInput Fp) input)
+
+/-- The wrap side's `finalize_other_proof` input at `r` rounds, by name, in the wrap
+layout's cell order: the claims, the evaluation block, `ft(ζω)`, the two previous-challenge
+vectors, and the digest last. No mask and no domain cell: the wrap side reads every
+accumulator and its domain is a constant. -/
+structure FopWrapInput (r : ℕ) (α : Type) where
```
martyall: r rounds of what?

